PR #3, opened 2022-07-15
week1-section2

simonkytr:
these are the exercises from section1-week1
exercise1
sw1

exercise2
sw2

exercise3
sw3

exercise4
sw4

exercise5
sw5

exercise6
sw6

exercise7
sw7

exercise8
sw8

exercise9
sw9

exercise10
sw10

### DIFF
--- a/section2-week1/excercise10/Main.cpp
+++ b/section2-week1/excercise10/Main.cpp
@@ -1,6 +1,6 @@
 /*Exercise 10 - Anything but Lowercase Vowel
 Rewrite the previous exercise but this time check that the given character is NOT a lowercase vowel. Use the same conditions you used previously and only change the relational operators. You are not allowed to use a negation operator to wrap the whole set of conditions.
-Hint: You might want to look up DeMorgan’s Law for this one.
+Hint: You might want to look up DeMorganâ€™s Law for this one.
 */
 
 #include <iostream>
@@ -20,14 +20,12 @@ int main()
 		{
 			cout << "This character is not a lowercase vowel" << endl;
 		}
-		cout << "This character is not a vowel" << endl;
+
 	}
 	else {
 		if (a == 'a' || a == 'e' || a == 'i' || a == 'o' || a == 'u')
 		{
 			cout << "This character is a lowercase vowel" << endl;
 		}
-		cout << "This character is not a vowel" << endl;
+		
 	}
-
-}
--- a/section2-week1/excercise10/Main.cpp
+++ b/section2-week1/excercise10/Main.cpp
@@ -15,17 +15,13 @@ int main()
 	cin >> a;
 
 	//identify the character
-	if (isupper(a)) {
-		if (a == 'a' || a == 'e' || a == 'i' || a == 'o' || a == 'u')
-		{
-			cout << "This character is not a lowercase vowel" << endl;
-		}
+	if (!(a == 'a' || a == 'e' || a == 'i' || a == 'o' || a == 'u'))
+	{
+		cout << "This character is not a lowercase vowel" << endl;
+	}
+
+	else {
+		cout << "This character is a lowercase vowel" << endl;
 
 	}
-	else {
-		if (a == 'a' || a == 'e' || a == 'i' || a == 'o' || a == 'u')
-		{
-			cout << "This character is a lowercase vowel" << endl;
-		}
-		
-	}
+}
--- a/section2-week1/excercise10/Main.cpp
+++ b/section2-week1/excercise10/Main.cpp
@@ -1,0 +1,33 @@
+/*Exercise 10 - Anything but Lowercase Vowel
+Rewrite the previous exercise but this time check that the given character is NOT a lowercase vowel. Use the same conditions you used previously and only change the relational operators. You are not allowed to use a negation operator to wrap the whole set of conditions.
+Hint: You might want to look up DeMorgan’s Law for this one.
+*/
+
+#include <iostream>
+using namespace std;
+
+int main()
+{
+	char a;
+
+	cout << "Do you know I can identify if your character is a lowercase vowel ? " << endl;
+	cout << "insert any character: ";
+	cin >> a;
+
+	//identify the character
+	if (isupper(a)) {
+		if (a == 'a' || a == 'e' || a == 'i' || a == 'o' || a == 'u')
+		{
+			cout << "This character is not a lowercase vowel" << endl;
+		}
+		cout << "This character is not a vowel" << endl;
+	}
+	else {
+		if (a == 'a' || a == 'e' || a == 'i' || a == 'o' || a == 'u')
+		{
+			cout << "This character is a lowercase vowel" << endl;
+		}
+		cout << "This character is not a vowel" << endl;
+	}
+
+}
--- a/section2-week1/exercise 8/Main.cpp
+++ b/section2-week1/exercise 8/Main.cpp
@@ -19,43 +19,24 @@ int main()
 		yes, no, maybe, perhaps
 	};
 	//inicialize variables
-	answers ans;
-	ans = yes;
+	answers ans = yes;
 
 
-	int v1;
+	int random;
 
 	//selecting the answer
 	srand(time(NULL));
-	v1 = rand() % (1 - 5);
+	random = rand() % (1 - 5);
 
 	char ask[100];
 	cout << "Would you like to know what happens in your future?" << endl;
 	cout << "\n+++++++++++++++++++++++++++++++++++++++++++++++++++" << endl;
 	cout << "\nYou can ask to me any question... " << endl;
 	cin >> ask;
-	
 
-	
+	ans = (answers)random;
+	cout << random << endl;
 
-	//giving and answer to the number random
-	switch (v1)
-	{
-	case 1:
-		ans = yes;
-		break;
-	case 2:
-		ans = no;
-		break;
-	case 3:
-		ans = maybe;
-		break;
-	case 4:
-		ans = perhaps;
-		break;
-	default:
-		break;
-	}
 	
 	//guessing the future
 	switch (ans)
@@ -77,3 +58,4 @@ int main()
 	}
 
 }
+
--- a/section2-week1/exercise 8/Main.cpp
+++ b/section2-week1/exercise 8/Main.cpp
@@ -26,7 +26,7 @@ int main()
 
 	//selecting the answer
 	srand(time(NULL));
-	random = rand() % (1 - 5);
+	random = rand() % 4;
 
 	char ask[100];
 	cout << "Would you like to know what happens in your future?" << endl;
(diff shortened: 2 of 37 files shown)


section2-week1/exercise5/Main.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <iostream>
+#include <stdio.h>      /* printf, scanf, puts, NULL */
+#include <stdlib.h>     /* srand, rand */
+#include <time.h>       /* time */
+using namespace std;
+
+int main()
+{
+	int coin, v1;
```
JDAlzate, line 14: What is the name v1 supposed to mean here?

section2-week1/exercise6/Main.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include <iostream>
+using namespace std;
+
+int main()
+{
+	cout << "WELCOME PLAYER, choose your character" << endl;
+	cout << "1 - Gardian\n2 - Wizard\n3 - Warrior\n4 - Healer" << endl;
+
+	int cho;
```
JDAlzate, line 13: What is cho supposed to mean? Is it choice? Is it really necessary to abbreviate choice?

JDAlzate: Remember to think about the reader when you write your code. What will make sense and be helpful for someone else reading your code?

section2-week1/exercise7/Main.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+#include <iostream>
+#include <stdio.h>      /* printf, scanf, puts, NULL */
+#include <stdlib.h>     /* srand, rand */
+#include <time.h>       /* time */
+using namespace std;
+
+int main()
+{
+	int value, v1, ans;
```
JDAlzate, line 14: Please change all the abbreviated names in your code to something more explicit

section2-week1/exercise7/Main.cpp
```diff
@@ -0,0 +1,103 @@
...
+	cout << "1 - Rock\n2 - Paper\n3 - Scissors" << endl;
+	cout << "choose you weapon: " << endl;
+
+	cin >> value;
+
+	cout << "\nI have selected the weapon: " << v1 << endl;
+
+
+	//comparing the weapons
+	switch (value)
```
JDAlzate, line 31: Maybe some enums would go along well with your switches and choices, instead of using ints directly,

section2-week1/exercise7/Main.cpp
```diff
@@ -0,0 +1,103 @@
...
+			break;
+		default:
+			cout << "You do not know how to play this game...";
+
+	}
+
+	//givin the beredict
+	switch (ans)
+	{
+	case 1:
```
JDAlzate, line 86: This is a nice idea, but an enum for this would also be useful.

section2-week1/exercise7/Main.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+	cout << "\nI have selected the weapon: " << v1 << endl;
+
+
+	//comparing the weapons
+	switch (value) 
+	{
+		case 1:
+			cout << "You have selected Rock" << endl;
+			if (v1 == 1)
```
JDAlzate, line 35: You're executing all your if's here. You could have used else if's to avoid this, but in this case, a switch would be a much better choice.

section2-week1/exercise9/Main.cpp
```diff
@@ -0,0 +1,27 @@
...
+int main()
+{
+	char a;
+
+	cout << "Do you know I can identify if your character is a lowercase vowel?" << endl;
+	cout << "insert any character: ";
+	cin >> a;
+
+	//identify the character
+	if ((islower(a)) && (a == 'a' || a == 'e' || a == 'i' || a == 'o' || a == 'u'))
```
JDAlzate, line 17: There is no need to check for `islower` here. You're already checking against lowercase vowels.

section2-week1/exercise 8/Main.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+
+int main()
+{
+	//enum variables
+	enum answers {
+		yes, no, maybe, perhaps
+	};
+	//inicialize variables
+	answers ans;
```
JDAlzate, line 22: Why have the declaration and the definition in different lines?

section2-week1/exercise 8/Main.cpp
```diff
@@ -0,0 +1,79 @@
...
+	cout << "Would you like to know what happens in your future?" << endl;
+	cout << "\n+++++++++++++++++++++++++++++++++++++++++++++++++++" << endl;
+	cout << "\nYou can ask to me any question... " << endl;
+	cin >> ask;
+	
+
+	
+
+	//giving and answer to the number random
+	switch (v1)
```
JDAlzate, line 42: There is no need for two variables and two switches here. You can compare v1 to the enum values directly.

section2-week1/excercise10/Main.cpp
```diff
@@ -0,0 +1,31 @@
...
+int main()
+{
+	char a;
+
+	cout << "Do you know I can identify if your character is a lowercase vowel ? " << endl;
+	cout << "insert any character: ";
+	cin >> a;
+
+	//identify the character
+	if (isupper(a)) {
```
JDAlzate, line 18: Remember what the instructions said. `Use the same conditions you used previously and only change the relational operators. You are not allowed to use a negation operator to wrap the whole set of conditions.` Also, this is not printing anything if the character is a non-vowel letter, which it should be doing

JDAlzate (changes requested):
Changes requested :)

section2-week1/exercise5/Main.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+	/* initialize random seed: */
+	srand(time(NULL));
+	v1 = rand() % (1 - 3);
```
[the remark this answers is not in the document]
simonkytr: I was reading the documentation about the use of random numbers with c++... and It explained that if you want a random number between (1-20) you have to type in this way

section2-week1/exercise6/Main.cpp
simonkytr: thanks! i'll take on mind :3

section2-week1/exercise3/Main.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+	cout << "do you want to prove my math skills? \n gimme a number and I say whether the number is a negative number, a positive number or cero " << endl;
+
+	cin >> number;
+
+	if (number < 0)
+	{
+		cout << "this number is negative " << endl;
+	}
+	if (number == 0)
```
JDAlzate, line 19: You're still checking if number is less than 0 AND if the number is 0, but these conditions could never be both true. You should instead use an `else if` here.

section2-week1/exercise7/Main.cpp
JDAlzate: You didn't change to an enum here.

section2-week1/exercise7/Main.cpp
JDAlzate: Or here

JDAlzate (changes requested):
Pending changes :)

section2-week1/exercise 8/Main.cpp
```diff
@@ -19,43 +19,24 @@ int main()
...
+	answers ans = yes;
 
 
-	int v1;
+	int random;
 
 	//selecting the answer
 	srand(time(NULL));
-	v1 = rand() % (1 - 5);
+	random = rand() % (1 - 5);
```
JDAlzate, line 29: I think this was commented before, but why are your using `1 - 5`? That's the same thing that `rand() % -4` so why not use that instead? But even so, `-4` is a bit unnecessarily specific, because you could just as well do `rand() % 4`

section2-week1/exercise7/Main.cpp
```diff
@@ -11,93 +11,96 @@ using namespace std;
 
 int main()
 {
-	int value, random, answer;
-	ans = 0;
+	int value, random;
```
JDAlzate, line 14: Why are you declaring these variables all the way up here if they are used later?

section2-week1/exercise7/Main.cpp
```diff
@@ -11,93 +11,96 @@ using namespace std;
...
 	cout << "1 - Rock\n2 - Paper\n3 - Scissors" << endl;
 	cout << "choose you weapon: " << endl;
 
-	cin >> value;
+	enum Weapon
+	{
+		Rock, Paper, Scissors
+	};
+	srand(time(NULL));
+	random = rand() % (1 - 4);
```
JDAlzate, line 26: Same comment here about `1 - 4`

JDAlzate, line 26: Random could have a more specific name. What _is_ random? If you read this variable name without any context, would you know what it is? You could specify that it's a weapon, for example. Something like `randomWeapon` or go even further and say `computerWeaponChoice` or `randomWeaponChoice`, something like that. What do you think?

section2-week1/exercise7/Main.cpp
```diff
@@ -11,93 +11,96 @@ using namespace std;
...
+	Weapon weapon = (Weapon)random;
 
-	cout << "\nI have selected the weapon: " << random << endl;
+	cin >> value;
```
JDAlzate, line 29: Same thing for the variable `value`. What is `value` supposed to mean? What about `userChoice` or `userWeapon`? Do you see how that would make the code a lot clearer?

section2-week1/exercise7/Main.cpp
```diff
@@ -11,93 +11,96 @@ using namespace std;
...
 
 
+	enum Answers
```
JDAlzate, line 32: Enums shouldn't really be named in plural. It may be a list of answers, but the actual variable type won't be. So for example, later you have the line `Answers answer = tie;` Which is weird that a single answer is of a type called Answers.

section2-week1/exercise7/Main.cpp
JDAlzate: Never mind, I see that this is just a random number that you're using to cast to a Weapon. However, `weapon` could also be more descriptive. Is it the user's weapon? Or the computer's weapon?

section2-week1/exercise7/Main.cpp
```diff
@@ -11,93 +11,96 @@ using namespace std;
...
-			{
-				answer = 1;
-			}
-			break;
-		default:
-			cout << "You do not know how to play this game...";
-
+	case Rock:
+		cout << "I have selected Rock" << endl;
+		if (value == 1)
```
JDAlzate, line 43: The variable `value` could be directly compared to the enum values, instead of the numbers. You could do `if (value == rock)` for instance. Also, remember to use `else` and `else if` when the values are mutually exclusive. Can `value` ever be 1 AND 2 at the same time?

JDAlzate (changes requested):
Comments! :)

PR #4, opened 2022-07-15
week1-section1 request

simonkytr:
screenshots week 1 section1
exercise 1
sw1
exercise 2
sw2
exercise3
sw3
exercise 4
sw4
exercise 6
sw5
exercise 7
sw7
exercise 8
sw8
exercise 9
sw9
exercise 10
sw10

### DIFF
--- a/exercise1/exercise1/main.cpp
+++ b/exercise1/exercise1/main.cpp
@@ -1,0 +1,12 @@
+#include <iostream>
+//Exercise 1 - Hello World
+//Write a program that will output “Hello World!” in the first line, then your nameand something about yourself in the second line
+
+
+int main()
+{
+	std::cout << "hello word¡" << std::endl;
+	std::cout << "my name is Diego and I'm so thrilled to start this internship" << std::endl;
+	std::cin.get();
+
+}
--- a/exercise1/exercise1/main.cpp
+++ b/exercise1/exercise1/main.cpp
@@ -1,11 +1,11 @@
 #include <iostream>
 //Exercise 1 - Hello World
-//Write a program that will output “Hello World!” in the first line, then your nameand something about yourself in the second line
+//Write a program that will output â€œHello World!â€ in the first line, then your nameand something about yourself in the second line
 
 
 int main()
 {
-	std::cout << "hello word¡" << std::endl;
+	std::cout << "Hello World!" << std::endl;
 	std::cout << "my name is Diego and I'm so thrilled to start this internship" << std::endl;
 	std::cin.get();
 
--- a/exercise1/exercise2/main.cpp
+++ b/exercise1/exercise2/main.cpp
@@ -1,0 +1,23 @@
+//Exercise 2 - Data Type Sizes
+Write a program that will output the size of the different data types in c++.
+Hint: You can add an indent in your output by using the \t escape character.
+
+#include <iostream>
+
+
+
+int main()
+{
+	std::cout << "Type data -- Data Size" << std::endl;
+	std::cout << "-------------------" << std::endl;
+	std::cout << " char -- " << sizeof(char) << std::endl;
+	std::cout << " int -- " << sizeof(int) << std::endl;
+	std::cout << " short -- " << sizeof(short) << std::endl;
+	std::cout << " float -- " << sizeof(float) << std::endl;
+	std::cout << " boolean -- " << sizeof(bool) << std::endl;
+	std::cout << " double -- " << sizeof(double) << std::endl;
+	std::cout << " long -- " << sizeof(long) << std::endl;
+
+	std::cin.get();
+}
+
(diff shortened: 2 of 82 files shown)


exercise1/exercise1/main.cpp
```diff
@@ -0,0 +1,12 @@
+#include <iostream>
+//Exercise 1 - Hello World
+//Write a program that will output �Hello World!� in the first line, then your nameand something about yourself in the second line
+
+
+int main()
+{
+	std::cout << "hello word�" << std::endl;
```
JDAlzate, line 8: Code is good, but you're not printing the correct text, be mindful of your code and the instructions :)

exercise1/exercise2/main.cpp
```diff
@@ -0,0 +1,23 @@
+//Exercise 2 - Data Type Sizes
+Write a program that will output the size of the different data types in c++.
```
JDAlzate, line 2: Don't upload code that doesn't compile. This lines aren't commented, so you code should not be able to run

exercise1/exercise3/main.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+	std::cout << "---------------------------------------------------" << std::endl;
+
+	std::cout << "Data Type  \t   ----  \tLower Limit \t ----  \tUpper Limit" << std::endl;
+
+	std::cout << "---------------------------------------------------" << std::endl;
+
+
+	Message("signed char", SCHAR_MIN, SCHAR_MAX);
+	Message("unigned char", -UCHAR_MAX / 2, UCHAR_MAX / 2);
```
JDAlzate, line 28: These values are not correct. Why are you making these divisions? What would be the min value of an unsigned that type (a data type that cant be negative)

exercise1/exercise4/main.cpp
```diff
@@ -0,0 +1,20 @@
...
+int main() {
+
+	int overflow = SHRT_MAX + 1;
+	int underflow = SHRT_MIN - 1;
+
+	std::cout << "��Esto pasa cuando no cuidas tus datos!!" << std::endl;
+
+	std::cout << "---------------------------------------" << std::endl;
+
+	std::cout << "El overflow sucede cuando usas una variable 'short' y el valor que utilizas es (" << SHRT_MAX << ")+1 \n El valor seria=" << overflow << std::endl;
```
JDAlzate, line 18: Test your code and make sure that it's working the way it should. There is no overflow or underflow going on here, ints can definitely store SHRT_MAX + 1 and SHRT_MIN - 1 without a problem.

exercise1/exercise7/main.cpp
```diff
@@ -0,0 +1,26 @@
...
+Write a program that will ask the user for three numbers (a,b,c) and output the following evaluations:
+a + 1 * b + 2 - c
+(a + 1) * (b + 2) - c
+*/
+#include <iostream>
+using namespace std;
+
+int main() 
+{
+	int a, b, c, x1, x2;
```
JDAlzate, line 11: There is no gain in declaring your variables before the rest of your code. Try to have them declared as close as possible or in the same line as the definition.

JDAlzate, line 11: Be mindful of your names. What are x1 and x2? Could they have more descriptive names?

exercise1/exercise8/Main.cpp
```diff
@@ -0,0 +1,27 @@
...
+*/
+#include <iostream>
+using namespace std;
+
+int main()
+{
+	float f,f2, c,c2;
+	cout << "insert a temperature in Fahrenheit ";
+	cin >> f;
+	c2 = (f - 32.0) * 5.0 / 9.0;
```
JDAlzate, line 15: Floats should use a .f at the end of their value. So 32.0 is a double, but 32.0f is a float.

exercise1/exercise9/Main.cpp
```diff
@@ -0,0 +1,40 @@
+/*Exercise 9 - Shopping
+Write a program that will simulate a shop where you can buy different items. Let the user choose how many items of each type they want, and then output the final price. Make sure to combine integer and floating-point types
+*/
+#include <iostream>
+using namespace std;
+
+int main() 
+{
+	int item1, item2, item3, price1, price2, price3, op1, op2, op3, suma;
```
JDAlzate, line 9: Same here, you don't need to declare your variables before everything else. There is no gain and in fact, the code is harder to read and follow.

JDAlzate (commented):
Review is up!

JDAlzate (changes requested):
Just one extra comment and you're good to go!

exercise1/exercise9/Main.cpp
JDAlzate: It seems like you're not taking into account all feedback. Please fix all comments before requesting a new review.

JDAlzate (changes requested):
Requesting all feedback being addressed

PR #6, opened 2022-07-18
Week1-section3

simonkytr:
I've finished section 3, ready for corrections
exercise1
sw1
exercise2
sw2
exercise3
sw3
exercise4
sw4
exercise5
sw5
exercise6
sw6
exercise7
sw7
exercise8
sw8
exercise9
sw9

### DIFF
--- a/week1-section3/exercise 3/Main.cpp
+++ b/week1-section3/exercise 3/Main.cpp
@@ -19,10 +19,11 @@ int main() {
 
 		if (i % 10 == 0) {
 			std::cout << i << ", ";
-			multiples = multiples + i;
+			multiples = + i;
+			continue;
 		}
-		sum = sum + i;//sum of the rest of the numbers
+		sum += i;//sum of the rest of the numbers
 	}
-	std::cout << std::endl << "Sum of the rest of numbers: "<< sum-multiples;//difference between the rest of the numbers and the multiples of 10
+	std::cout << std::endl << "Sum of the rest of numbers: "<< sum;//difference between the rest of the numbers and the multiples of 10
 	std::cin.get();
 }
--- a/week1-section3/exercise 3/Main.cpp
+++ b/week1-section3/exercise 3/Main.cpp
@@ -1,0 +1,28 @@
+/*Exercise 3 - Multiples of 10
+Write a program that will output multiples of 10 between 1 and 100. Use a for-loop that goes through numbers 1-100.	
+
+At the end of the program, output the sum of all the numbers that were NOT multiples of 10. Do NOT use an else statement for this program and do NOT use more than one loop.
+
+*/
+#include <iostream>
+
+int main() {
+
+	
+	int sum=0;
+	int multiples=0;
+	std::cout << "These the multyples of 100" << std::endl;
+
+	for (int i = 0; i <= 100; i++)
+	{
+		//give the multiples of 10
+
+		if (i % 10 == 0) {
+			std::cout << i << ", ";
+			multiples = multiples + i;
+		}
+		sum = sum + i;//sum of the rest of the numbers
+	}
+	std::cout << std::endl << "Sum of the rest of numbers: "<< sum-multiples;//difference between the rest of the numbers and the multiples of 10
+	std::cin.get();
+}
--- a/week1-section3/exercise 4/Main.cpp
+++ b/week1-section3/exercise 4/Main.cpp
@@ -9,16 +9,13 @@ This time you can only use one for loop and it should not loop more than 10 time
 
 int main() {
 
-	
-	int multiples;
 	std::cout << "These are the multyples of 100" << std::endl;
 
-	for (int i = 10; i < 100; i=i+10)
+	for (int i = 10; i <= 100; i+=10)
 	{
 		//give the multiples of 10
 
-		multiples = i;
-		std::cout << multiples << ", ";
+		std::cout << i << ", ";
 
 	}
 	
--- a/week1-section3/exercise 4/Main.cpp
+++ b/week1-section3/exercise 4/Main.cpp
@@ -1,0 +1,25 @@
+/*
+Exercise 4 - More Multiples of 10
+Rewrite the previous program, but this time loop just outputs the multiples of 10, no need to output the sum of the rest of the numbers.
+
+This time you can only use one for loop and it should not loop more than 10 times.
+*/
+
+#include <iostream>
+
+int main() {
+
+	
+	int multiples;
+	std::cout << "These are the multyples of 100" << std::endl;
+
+	for (int i = 10; i < 100; i=i+10)
+	{
+		//give the multiples of 10
+
+		multiples = i;
+		std::cout << multiples << ", ";
+
+	}
+	
+}
--- a/week1-section3/exercise 5/Main.cpp
+++ b/week1-section3/exercise 5/Main.cpp
@@ -15,25 +15,30 @@ int main(){
 	int  insert, sum=0;
 	int size = 0;
 
-
-
 	for ( int i = 0; i < 10; i++){
-		std::cout << "insert any NUMBER: ";
-		std::cin >> insert;
-		sum = sum + insert;
-		insert = numbers[i];
-		
+
+
 		//Error: interrup the loop
-		
-			if (std::cin.fail() || std::cin.bad()) {
+		if (std::cin.fail() || std::cin.bad()) {
 			std::cin.clear();
 			std::cin.ignore(100, '\n');
 
-			std::cout << "\ntotal number entered: " << 1+i << std::endl;
+			std::cout << "\ntotal number entered: " << size << std::endl;
 			std::cout << "\nTotal sum of numbers entered: " << sum << std::endl;
 
 			return 0;
 		}
+		
+
+		std::cout << "insert any NUMBER: ";
+		std::cin >> insert;
+		numbers[i] = insert;
+		sum  += insert;
+		
+		
+		size++;
+		
+			
 
 	}
 
--- a/week1-section3/exercise 5/Main.cpp
+++ b/week1-section3/exercise 5/Main.cpp
@@ -1,0 +1,46 @@
+/*Exercise 5 - Arrays
+Create an array of whatever size you want. Let the user input as many numbers as they want and populate the array with them. If they enter any invalid input, output the whole array of numbers and the sum of them all and the total numbers they input.
+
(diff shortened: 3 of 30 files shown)


week1-section3/exercise1/Main.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+	std::cout << "Can you guess the number that I am thinking??\n++++++++++++++++++++++\n";
+	actual = 5;//value to guess
+
+	while ( i < 5 )
+	{
+		std::cout << "\nPlease enter a number between 1 - 10: ";
+		std::cin >> number;
+		
+		if (number == actual && !(i >= 3)) {
```
JDAlzate, line 21: This is perfectly fine logic, but try to avoid wrapping an expression with a negation operator (`!`) if it is easily negated without it. In this case, instead of `!(i >= 3)`, you could write `i < 3`.

week1-section3/exercise1/Main.cpp
```diff
@@ -0,0 +1,41 @@
...
+			}
+			
+		else {
+			if (i == 4)
+			{
+				std::cout << "\nYou've lost your attempts. I'm sorry... LOSER!\n";//annoying message
+			}
+			//continue guessing
+			i++;
+			continue;
```
JDAlzate, line 38: No need to add continue at the end of the loop. It will naturally continue anyway.

week1-section3/exercise3/Main.cpp
```diff
@@ -0,0 +1,83 @@
+/*Exercise 2 - More Input Validation
```
JDAlzate, line 1: Same comment about naming your projects. This is exercise 2 but the project is called exercise3.

week1-section3/exercise3/Main.cpp
```diff
@@ -0,0 +1,83 @@
...
+	std::cout << "Welcome to this quiz\How many doy you know about general culture?"<<std::endl;
+	std::cout << "Remember to use F to false or T to true\nLet's play!!\n"<<std::endl;
+
+
+	//creating asks and answers
+	std::string question[5] = {"The paramecium is a flower? ","does the platypus lay eggs? ", "Is the whale a mammal? ","Is the bat a bird? ", "Is the tomatoe a fruit? "};
+	std::string answer[5] = {"F","T","T","F","T"};
+	
+	//creating inputs
+	std::string answer2;
```
JDAlzate, line 18: Choose the names of your variables carefully. Don't just add a 2 to a name that is already in use, especially in this case where these aren't related at all.

Having a variable called `answer` and another called `answer2` makes it sound like there are two answers. Nothing about it makes me think that `answer` is an array of correct answers, and `answer2` is the actual input of the user.

week1-section3/exercise3/Main.cpp
```diff
@@ -0,0 +1,83 @@
...
+	int score=0;
+
+	int ansWrong=0;//to validate the wrong typing
```
JDAlzate, line 21: `ansWrong` is a bit undescriptive. I don't understand what it's supposed to mean from just reading it. Is it some kind of value for a wrong answer? Is it the count of wrong answers?

week1-section3/exercise3/Main.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+	for ( int i = 0; i < 5; i++)
```
JDAlzate, line 23: Let's discuss this exercise tomorrow :) It seems to be working fine for the most part, but it is a bit over-complicated.

week1-section3/exercise 3/Main.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+	for (int i = 0; i <= 100; i++)
+	{
+		//give the multiples of 10
+
+		if (i % 10 == 0) {
+			std::cout << i << ", ";
+			multiples = multiples + i;
+		}
+		sum = sum + i;//sum of the rest of the numbers
```
JDAlzate, line 24: Why add the multiples as well if you are going to subtract it in the end anyway. Maybe there's a way to simply skip the sum of multiples without needing to use an else statement?

week1-section3/exercise 3/Main.cpp
```diff
@@ -0,0 +1,28 @@
...
+	int multiples=0;
+	std::cout << "These the multyples of 100" << std::endl;
+
+	for (int i = 0; i <= 100; i++)
+	{
+		//give the multiples of 10
+
+		if (i % 10 == 0) {
+			std::cout << i << ", ";
+			multiples = multiples + i;
```
JDAlzate, line 22: This is fine and there is no need to change it. But just as a comment, keep in mind that you can also write `multiples += i`

week1-section3/exercise 4/Main.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include <iostream>
+
+int main() {
+
+	
+	int multiples;
+	std::cout << "These are the multyples of 100" << std::endl;
+
+	for (int i = 10; i < 100; i=i+10)
```
JDAlzate, line 16: Same as previous comment. Keep in mind you could write `i += 10`.

week1-section3/exercise 4/Main.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+This time you can only use one for loop and it should not loop more than 10 times.
+*/
+
+#include <iostream>
+
+int main() {
+
+	
+	int multiples;
```
JDAlzate, line 13: Why do you need this variable? Why not use `i` directly?

week1-section3/exercise 4/Main.cpp
```diff
@@ -0,0 +1,25 @@
...
+	std::cout << "These are the multyples of 100" << std::endl;
+
+	for (int i = 10; i < 100; i=i+10)
+	{
+		//give the multiples of 10
+
+		multiples = i;
+		std::cout << multiples << ", ";
```
JDAlzate, line 21: Are you sure this is working correctly? Is 100 actually printing?

week1-section3/exercise 5/Main.cpp
```diff
@@ -0,0 +1,46 @@
...
+*/
+
+#include <iostream>
+
+
+
+
+int main(){
+
+	int numbers[10];
```
JDAlzate, line 14: Remember that there is no need to declare all your variables before your logic. Declare your variables as close to their definition and as close to their use as possible.

week1-section3/exercise 5/Main.cpp
```diff
@@ -0,0 +1,46 @@
...
+	int  insert, sum=0;
+	int size = 0;
+
+
+
+	for ( int i = 0; i < 10; i++){
+		std::cout << "insert any NUMBER: ";
+		std::cin >> insert;
+		sum = sum + insert;
```
JDAlzate, line 23: Should you really be doing this before you validate the input? You're adding random values to your sum.

week1-section3/exercise 5/Main.cpp
```diff
@@ -0,0 +1,46 @@
...
+		insert = numbers[i];
```
JDAlzate, line 24: Remember that you want to have the variable you want to assign a value to BEFORE the `=` and the value you want to assign to it AFTER the `=`. 

This should be written as `numbers[i] = insert;`

week1-section3/exercise 6/Main.cpp
```diff
@@ -0,0 +1,62 @@
...
+Hint: You�ll need to look up stuff here, too. How would you detect invalid input and an enter key press.
+*/
+#include<iostream>
+#include<conio.h>
+
+
+int main()
+{
+
+	int numbers[10];
```
JDAlzate, line 17: Let's review this exercise tomorrow as well

week1-section3/exercise/Main.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+	std::cout << "Welcome to the lottery!" << std::endl;
+	
+	int arrayInput[5];
+	
+	for (int i = 0; i < 5; i++)
+	{
+
+		std::cout << "Please enter entry a number between (1-20): ";
+		std::cin >> arrayInput[i];
```
JDAlzate, line 23: Remember you're supposed to handle invalid input.

week1-section3/exercise 8/Main.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <cstdlib>
+
+
+int main()
+{
+
+
+	std::cout << "Welcome to the lottery!" << std::endl;
+
```
JDAlzate, line 15: Is it really necessary to add all this empty space between the lines of your code?

week1-section3/exercise 8/Main.cpp
```diff
@@ -0,0 +1,56 @@
...
+	
+
+
+
+	int arrayInput[5];
+	int input;
```
JDAlzate, line 21: Doesn't seem like this variable is used for anything is it?

week1-section3/exercise 8/Main.cpp
```diff
@@ -0,0 +1,56 @@
...
+	std::cout << "Lottery results: ";
+	int arrayRandom[5];
+	for (int i = 0; i < 5; i++)
+	{
+		/* initialize random seed: */
+
+		arrayRandom[i] = 1 + rand() % 20;
+		for (int j = 0; j < 5; j++)
+		{
+			if (arrayInput[i] == arrayRandom[j]) prize+10;
```
JDAlzate, line 44: It doesn't seem like this line is doing anything here 🤔

JDAlzate: You're also checking your values before completely initializing your `arrayRandom` values.

week1-section3/exercise 9/Main.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+	}
+	
+	wordReverse[size] = 0;
+	std::cout << "This string reversed is: " << wordReverse << std::endl;
+
+
+	//------------------------------------------------------------------------
+
+	char word2[30] = {"D i e g o"};
```
JDAlzate, line 43: You should be using single characters here. "D i e g o" is still a string, since you're using double quotes. Remember a character is defined using single quotes such as 'D'.

JDAlzate (changes requested):
Review is up!

JDAlzate (approved):
Approved ignoring all comments about coding standards, since I understand that this review was made before the standards were sent.

PR #8, opened 2022-07-26
Week2 section 2

simonkytr:
these are the exercises from the section 2, week 2
exercise 2
2w2s2
exercise 3
3w2s2
exercise 4
4w2s2
exercise 4.1
4 2w2s2
exercise 4.2
4 3w2s2
4 3 1w2s2

### DIFF
--- a/week2-section2/exercise2/Main.h
+++ b/week2-section2/exercise2/Main.h
@@ -1,0 +1,2 @@
+#pragma once
+
--- a/week2-section2/exercise2/PlayList.cpp
+++ b/week2-section2/exercise2/PlayList.cpp
@@ -5,55 +5,57 @@
 	FPlaylist::FPlaylist()
 	{
 		PlaylistTitle = "";
-		Songs[MaxSongCount];
 		songCount = 0;
 	}
 	FPlaylist::FPlaylist ( const std::string InPlayListTitle, const int InSongCount )
 	{
 		PlaylistTitle = InPlayListTitle;
-		Songs[MaxSongCount];
 		songCount = InSongCount;
 	}
 	 bool FPlaylist::bAddSong (const FSong InSong ) 
 	{
-		 for (int i = 0; i < MaxSongCount; i++)
+		 for (int i = 0; i < songCount; i++)
 		 {
-			 if (InSong.bIsValid())
+			 if (Songs[i].GetArtistName() == "")
 			 {
-				 std::cout << "Song Added!" << i << std::endl;
-				 Songs[i];
+				 std::cout << "Song #" << 1 + i << " Added!" << std::endl;
+				 Songs[i] = InSong;
 				 return true;
-			 }
-			 else {
-				 std::cout << "Invalid Song" << std::endl;
-				 return false;
 			 }
 		 }
 	};
 	bool FPlaylist::bPlaylistValid() const
 	{
-		if (PlaylistTitle != "" && songCount <= 1)
+		if ((PlaylistTitle != "") && (songCount <= 1) && (songCount < MaxSongCount))
 		{
 			return true;
 		}
 		else
 		{
-			std::cout << "Invalid Playlist" << std::endl;
 			return false;
 		}
 	}
 	FSong FPlaylist::GetSong (int SongIndex) const
 	{
-		for (SongIndex = 0; SongIndex < MaxSongCount; SongIndex++)
+		for (int OutSongIndex = 0; OutSongIndex < songCount; OutSongIndex++)
 		{
-			return Songs[SongIndex];
+			if (SongIndex == OutSongIndex) {
+				return Songs[SongIndex];
+			}
 		}
 	};
-	void FPlaylist::displayPlaylist()
+	void FPlaylist::DisplayPlaylist() const
 	{
+		std::cout << "\nPlaylist: " << PlaylistTitle << std::endl;
 		std::cout << "\nSong Title" << "\tArtist Name" << "\tDuration" << std::endl;
-		for (int i = 0; i < MaxSongCount; i++)
+		std::cout << "-----------------------------------------------------------------" << std::endl;
+		for (int i = 0; i < songCount; i++)
 		{
 			std::cout << Songs[i].GetSongTitle() << "\t" << Songs[i].GetArtistName() << "\t" << Songs[i].GetFormatedDuration() << std::endl;
 		}
 	};
+	std::string FPlaylist::GetPlaylistTitle() const
+	{
+		return  PlaylistTitle;
+	};
+	
--- a/week2-section2/exercise2/PlayList.cpp
+++ b/week2-section2/exercise2/PlayList.cpp
@@ -1,0 +1,59 @@
+#include <iostream>
+#include "Song.h"
+#include"PlayList.h"
+
+	FPlaylist::FPlaylist()
+	{
+		PlaylistTitle = "";
+		Songs[MaxSongCount];
+		songCount = 0;
+	}
+	FPlaylist::FPlaylist ( const std::string InPlayListTitle, const int InSongCount )
+	{
+		PlaylistTitle = InPlayListTitle;
+		Songs[MaxSongCount];
+		songCount = InSongCount;
+	}
+	 bool FPlaylist::bAddSong (const FSong InSong ) 
+	{
+		 for (int i = 0; i < MaxSongCount; i++)
+		 {
+			 if (InSong.bIsValid())
+			 {
+				 std::cout << "Song Added!" << i << std::endl;
+				 Songs[i];
+				 return true;
+			 }
+			 else {
+				 std::cout << "Invalid Song" << std::endl;
+				 return false;
+			 }
+		 }
+	};
+	bool FPlaylist::bPlaylistValid() const
+	{
+		if (PlaylistTitle != "" && songCount <= 1)
+		{
+			return true;
+		}
+		else
+		{
+			std::cout << "Invalid Playlist" << std::endl;
+			return false;
+		}
+	}
+	FSong FPlaylist::GetSong (int SongIndex) const
+	{
+		for (SongIndex = 0; SongIndex < MaxSongCount; SongIndex++)
+		{
+			return Songs[SongIndex];
+		}
+	};
+	void FPlaylist::displayPlaylist()
+	{
+		std::cout << "\nSong Title" << "\tArtist Name" << "\tDuration" << std::endl;
+		for (int i = 0; i < MaxSongCount; i++)
+		{
+			std::cout << Songs[i].GetSongTitle() << "\t" << Songs[i].GetArtistName() << "\t" << Songs[i].GetFormatedDuration() << std::endl;
+		}
+	};
--- a/week2-section2/exercise2/PlaylistContainer.cpp
+++ b/week2-section2/exercise2/PlaylistContainer.cpp
@@ -2,54 +2,56 @@
 #include "PlaylistContainer.h"
 #include "PlayList.h"
 
-FPlaylistContainer::FPlaylistContainer()
-{
-	Playlists[MaxPlaylistCount];
-	PlaylistCount = 0;
-};
-
-FPlaylistContainer::FPlaylistContainer ( const int InPlaylistCount)
-{
-	Playlists[MaxPlaylistCount];
-	PlaylistCount = InPlaylistCount;
-};
-
-bool FPlaylistContainer::bAddPlaylist (const FPlaylist InPlaylist) 
-{
-	for (int i = 0; i < MaxPlaylistCount; i++)
+	FPlaylistContainer::FPlaylistContainer()
 	{
-		if (InPlaylist.bPlaylistValid())
+		Playlists[MaxPlaylistCount];
+		PlaylistCount = 0;
+	};
+
+	FPlaylistContainer::FPlaylistContainer ( const int InPlaylistCount)
+	{
+		Playlists[MaxPlaylistCount];
+		PlaylistCount = InPlaylistCount;
+	};
+
+	bool FPlaylistContainer::bAddPlaylist (const FPlaylist InPlaylist) 
+	{
+		for (int i = 0; i < PlaylistCount; i++)
 		{
-			std::cout << "PlayList Added!" << std::endl;
-			return true;
+			if (Playlists[i].GetPlaylistTitle() == "")
+			{
+				std::cout << "PlayList Added! " << i << std::endl;
+				Playlists[i] = InPlaylist;
+				return true;
+			}
 		}
-		else {
-			std::cout << "Invalid PlayList" << std::endl;
-			return false;
-		}
-	}
-};
+	};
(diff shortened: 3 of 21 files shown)


week2-section2/exercise2/Song.cpp
```diff
@@ -0,0 +1,53 @@
...
+			return std::to_string(Minutes) + ":" + std::to_string(Seconds);
+		}
+		else
+		{
+			return std::to_string(Minutes) + ":" +"0" + std::to_string(Seconds);
+		}
+		
+	};
+
+	bool FSong::bIsValid() const
```
JDAlzate, line 44: Functions don't need the b prefix. That's only for bool variables. The only rule for bool functions is for them to answer a true or false question, which IsValid does correctly :)

week2-section2/exercise2/Song.cpp
```diff
@@ -0,0 +1,53 @@
...
+	{
+		if (( Duration >= 0 ) && ( ArtistName != "" ) && ( SongTitle != "" ))
+		{
+			return true;
+		}
+		else {
```
JDAlzate, line 50: Remember the braces syntax.

week2-section2/exercise2/Song.cpp
```diff
@@ -0,0 +1,53 @@
...
+	};
+
+	std::string FSong::GetArtistName() const
+	{
+		return ArtistName;
+	};
+	
+	std::string FSong::GetFormatedDuration() const
+	{
+		int Minutes = Duration / 60;
```
JDAlzate, line 31: Minutes and seconds could be const.

week2-section2/exercise2/Song.cpp
```diff
@@ -0,0 +1,53 @@
...
+		else
+		{
+			return std::to_string(Minutes) + ":" +"0" + std::to_string(Seconds);
+		}
+		
+	};
+
+	bool FSong::bIsValid() const
+	{
+		if (( Duration >= 0 ) && ( ArtistName != "" ) && ( SongTitle != "" ))
```
JDAlzate, line 46: While
```
if (bCondition)
{
    return true;
}
else
{
    return false;
}
```
works just fine, it's much more complex than it needs to be. Simply writing:
```
return Duration >= 0 && ArtistName != "" && SongTitle != "";
```
is enough and easier to read.

JDAlzate: Also, is a song that lasts 0 seconds really valid?

week2-section2/exercise2/PlayList.cpp
```diff
@@ -0,0 +1,61 @@
+#include <iostream>
+#include "Song.h"
+#include"PlayList.h"
```
JDAlzate, line 3: Since you are including "Playlist.h" and "Playlist.h" already includes <iostream> and "Song.h", there is no need to include them here as well.

week2-section2/exercise2/PlayList.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+	FPlaylist::FPlaylist()
+	{
+		PlaylistTitle = "";
+		songCount = 0;
```
JDAlzate, line 8: Remember the conventions.

week2-section2/exercise2/PlayList.cpp
```diff
@@ -0,0 +1,61 @@
...
+	}
+	FPlaylist::FPlaylist ( const std::string InPlayListTitle, const int InSongCount )
+	{
+		PlaylistTitle = InPlayListTitle;
+		songCount = InSongCount;
```
JDAlzate, line 13: The goal of SongCount is to keep track of the number of songs currently in the playlist, not to specify some max value of songs (which is what MaxSongCount is for).

That means that this constructor should not take in a parameter to initialize this value, since it should always start as 0 and increase with each added song.

week2-section2/exercise2/PlayList.cpp
```diff
@@ -0,0 +1,61 @@
...
+	}
+	 bool FPlaylist::bAddSong (const FSong InSong ) 
+	{
+		 for (int i = 0; i < songCount; i++)
```
JDAlzate, line 17: With what I commented above, instead of looping through the current songs and using the `SongCount` as the max number of songs, this method could instead use SongCount as an index to add the song if there is still more space for new songs.

This way, you don't have to loop every time. That is why we have the SongCount variable in the first place.

JDAlzate (changes requested):
Please fix the current comments for all the program (there are some other places where the `if (condition) return true; else return false;` block of code occurs) and fix the Count issue for playlists and playlists containers before requesting a re-review :).

week2-section2/exercise2/Spotify.cpp
```diff
@@ -0,0 +1,181 @@
...
+	ActiveMenuCommand = EMenuCommand::MainMenu;
+	PlaylistContainer = FPlaylistContainer(0);
+};
+
+//------------------------------------------------------
+void FSpotify::OpenMainMenu() 
+{
+	std::system("cls");
+	std::cout << "Welcome to Spotify! What would you like to do?\n1 - Create a Playlist\n2 - Browse Playlists\n3 - Exit" << std::endl;
+	int InputConverter;
```
JDAlzate, line 16: This isn't really an `InputConverter` it's just the user's input. The name should describe what the variable _is_, not what it is possibly _used for_. So `UserInput` should be more than enough :)

week2-section2/exercise2/Spotify.cpp
```diff
@@ -0,0 +1,181 @@
...
+			std::cin.clear();
+			std::cin.ignore(100, '\n');
+			std::cout << "Invalid Option, Insert a valid option" << std::endl;
+			std::cout << "Welcome to Spotify! What would you like to do?\n1 - Create a Playlist\n2 - Browse Playlists\n3 - Exit" << std::endl;
+			std::cin >> InputConverter;
+	}
+
+	ActiveMenuCommand = (EMenuCommand)InputConverter;
+	std::system("cls");
+	FSpotify::RunApp();
```
JDAlzate, line 30: Two things here:
1. No need for the `FSpotify::`, you're already within the FSpotify scope.
2. Is this function call really necessary? Think about what RunApp does.

week2-section2/exercise2/Spotify.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+//--------------------------------------------------------
+void  FSpotify::OpenCreatePlaylistMenu() 
+{
+	//Creating the playlist
+	std::system("cls");
+	std::cout << "Please enter a name for the playlist: " << std::endl;
+	char TitlePlaylist[500];
+	std::cin.getline(TitlePlaylist, 500,'\n');
+	std::cin >> TitlePlaylist;
```
JDAlzate, line 41: You're getting the whole line and assigning it to the TitlePlaylist char array and then you're taking the input again and assigning it again to the TitlePlaylist. 

There is no need to call std::cin >> TitlePlaylist if you've already done a getline.

week2-section2/exercise2/Spotify.cpp
```diff
@@ -0,0 +1,181 @@
...
+	FPlaylist Playlist = FPlaylist(TitlePlaylist, 0);
+	
+	std::cout << "Let's add a song to " << TitlePlaylist << std::endl;
+	//Creating the song loop
+
+	bool bWantToAddASong = true;
+	while (bWantToAddASong == true)
```
JDAlzate, line 48: There isn't really a need to check if some `bool == true`. It's the same thing as just using the bool as a condition itself.

So `while(bWantToAddSong)` is enough :).

week2-section2/exercise2/Spotify.cpp
```diff
@@ -0,0 +1,181 @@
...
+			std::cin >> MenuSelection;
+		}
+		if (MenuSelection == 2)
+		{
+			//adding the playlist in container
+			PlaylistContainer.AddPlaylist(Playlist);
+			bWantToAddASong = false;
+			FSpotify::OpenMainMenu();
+		}
+		else if (MenuSelection == 1) {
```
JDAlzate, line 96: This might be a small comment, but I find it a little misleading to have the `MenuSelection == 2` as the first condition and `MenuSelection == 1` as the second condition. I would've expected it to be the other way around to keep the number order, what do you think?

JDAlzate, line 96: Keep in mind that after you exit the while loop, your only options are 1 or 2, so there is no need for an `else if`, you could just use an `else`.

week2-section2/exercise2/Spotify.cpp
```diff
@@ -0,0 +1,181 @@
...
+			std::cout << "Invalid Option, Insert a valid option" << std::endl;
+			std::cout << "1 - Enter another song\n2 - Go back to main menu" << std::endl;
+			std::cin >> MenuSelection;
+		}
+		if (MenuSelection == 2)
+		{
+			//adding the playlist in container
+			PlaylistContainer.AddPlaylist(Playlist);
+			bWantToAddASong = false;
+			FSpotify::OpenMainMenu();
```
JDAlzate, line 94: Think about the design of the program. The way we switch between menus is using the ActiveMenuCommand, not calling the menu directly. You're opening the menu here (which shouldn't be necessary) and you're never changing the active menu command.

week2-section2/exercise2/Spotify.cpp
```diff
@@ -0,0 +1,181 @@
...
+};
+
+//----------------------------------------------------------
+void FSpotify::OpenBrowsePlaylistsMenu() 
+{
+	std::system("cls");
+	std::cout << "Here are your playlist(s):" << std::endl;
+	
+	//Looking the playlists saved
+	PlaylistContainer.GetCurrentPlaylistCount();
```
JDAlzate, line 110: Your getter should not be the one responsible for printing the playlist count. A function called `GetCurrentPlaylistCount` should only be returning the current playlist count and nothing else. If you want to print that, you'd have to do it here or create a separate method for printing the current playlist count.

week2-section2/exercise2/Spotify.cpp
```diff
@@ -0,0 +1,181 @@
...
+		std::cout << "Here are your playlist(s):" << std::endl;
+
+		//Looking the playlists saved
+		PlaylistContainer.GetCurrentPlaylistCount();
+		std::cout << "\nWhich playlist do you want to see: " << std::endl;
+
+		std::cin >> InputConverter;
+	}
+	std::system("cls");
+	PlaylistContainer.GetPlaylist(InputConverter - 1);
```
JDAlzate, line 132: Same here. Please fix this for any getter that is printing something instead of just returning a value.

week2-section2/exercise2/Spotify.cpp
```diff
@@ -0,0 +1,181 @@
...
+		std::cin.clear();
+		std::cin.ignore(100, '\n');
+		std::cout << "Invalid Option, Insert a valid option" << std::endl;
+		std::cout << "1 - Pick another playlist\n2 - Go back to main menu" << std::endl;
+		std::cin >> MenuSelection;
+	}
+	if (MenuSelection == 1)
+	{
+		std::system("cls");
+		FSpotify::OpenBrowsePlaylistsMenu();
```
JDAlzate, line 150: And please fix these open calls. With the design of your program, you shouldn't be opening the menus directly.

week2-section2/exercise2/Main.h
```diff
@@ -0,0 +1,2 @@
+#pragma once
```
JDAlzate, line 1: Not all cpp files need a .h file. This is mainly for class declarations.

week2-section2/exercise2/Spotify.cpp
```diff
@@ -0,0 +1,181 @@
...
+			FSpotify::OpenCreatePlaylistMenu();
+		}
+		else if (ActiveMenuCommand == FSpotify::EMenuCommand::BrowsePlaylists)
+		{
+			FSpotify::OpenBrowsePlaylistsMenu();
+		}
+	}
+	if (ActiveMenuCommand == FSpotify::EMenuCommand::Exit)
+	{
+		return 0;
```
JDAlzate, line 179: Why does RunApp return an int? It's always returning zero, what would that mean?

week2-section2/exercise2/Spotify.cpp
```diff
@@ -0,0 +1,181 @@
...
+		else if (ActiveMenuCommand == FSpotify::EMenuCommand::CreatePlaylist)
+		{
+			FSpotify::OpenCreatePlaylistMenu();
+		}
+		else if (ActiveMenuCommand == FSpotify::EMenuCommand::BrowsePlaylists)
+		{
+			FSpotify::OpenBrowsePlaylistsMenu();
+		}
+	}
+	if (ActiveMenuCommand == FSpotify::EMenuCommand::Exit)
```
JDAlzate, line 177: Also, you wouldn't need an if for this. Remember that this part of the code (after the while loop) is only executed if the condition in the while loop fails, so this wouldn't be executed if ActiveMenuCommand wasn't Exit.

JDAlzate: Also, please remove the unnecessary `FSpotify::` uses in your code.

JDAlzate (changes requested):
Review updated! :)

week2-section2/exercise2/PlayList.cpp
```diff
@@ -0,0 +1,56 @@
...
+		SongCount = 0;
+	}
+
+	FPlaylist::FPlaylist(const std::string InPlayListTitle, const int InSongCount)
+	{
+		PlaylistTitle = InPlayListTitle;
+		SongCount = InSongCount;
+	}
+
+	bool FPlaylist::AddSong(const FSong InSong )
```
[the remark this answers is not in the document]
JDAlzate: You didn't address this

week2-section2/exercise2/PlaylistContainer.cpp
```diff
@@ -28,10 +23,11 @@
 		}
 	};
 
-	FPlaylist FPlaylistContainer::GetPlaylist ( int InPlaylistIndex) const
+	//------------------------------------------------------------
+	FPlaylist FPlaylistContainer::GetPlaylist(int InPlaylistIndex) const
 	{
 		std::cout << "\nResults: " << std::endl;
-		if (Playlists[InPlaylistIndex].GetPlaylistTitle() != "")
+		if (Playlists[InPlaylistIndex].IsPlaylistValid()  == true)
```
JDAlzate, line 30: Remember that the `== true` is not necessary. You can simply use the result of the method itself since it returns a boolean.

week2-section2/exercise2/Spotify.cpp
JDAlzate: You didn't address this either 👀

JDAlzate (changes requested):
Looks a lot better! Just a few comments more.

JDAlzate (approved):
Approved! ✅

PR #9, opened 2022-08-11
SteamReview

simonkytr:


GaxysGalactic (changes requested):
Very well done! Functionality looks great, my main comments are on code formatting and some general suggestions. There could be some issues with the input validation with the copying thing, so maybe check that first. Some other positive points:

- Great job with function overloading, especially with InputValidate!
- Encapsulation in the Steam class (private functions) is important, glad you caught that.

SteamReview/SteamReview/CategoryContainer.cpp
```diff
@@ -0,0 +1,64 @@
...
+		CategoryCount++;
+		return true;
+	}
+	return false;
+};
+
+//---------------------------------------------------------
+bool FCategoryContainer::DeleteCategory(const int InCategoryIndex)
+{
+	if ((InCategoryIndex >= 0))
```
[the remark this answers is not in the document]
JDAlzate: Completely agree. @simonkytr it's important always to have a _reason_ to do something. If there is no reason to have these parentheses, it is just noise in your code, decreasing readability.

SteamReview/SteamReview/CategoryContainer.h
```diff
@@ -0,0 +1,19 @@
...
+	FCategory Categories[MaxCategoryCount];
+	int CategoryCount;
+public:
+	FCategoryContainer();
+	bool AddCategory(const FCategory InCategory);
+	bool DeleteCategory(const int InCategoryIndex);
+	FCategory GetCategory(const int InCategoryIndex) const;
+	bool AddGameToCategory(const int IndexCategory, const FGame InGame);
+	void PrintCategory() const;
+	int GetCurrentCategory() const;
```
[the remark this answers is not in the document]
JDAlzate: Nice! It's very useful to differentiate between the actual object and a count variable. I agree that `GetCurrentCategory()` is not an appropriate name, since it sounds as if it would return a category rather than an int.

SteamReview/SteamReview/Steam.cpp
```diff
@@ -0,0 +1,320 @@
...
+	std::cout << "----------------------------------------" << std::endl;
+	std::cout << "1 - Create a new category" << std::endl;
+	std::cout << "2 - Delete a category" << std::endl;
+	std::cout << "3 - Main manu" << std::endl;
+	std::cout << "----------------------------------------" << std::endl;
+	std::cout << "choose an option" << std::endl;
+
+	int UserInput;
+	std::cin >> UserInput;
+	InputValidate(UserInput, 0, 4);
```
JDAlzate, line 87: Why are you validating this with the range 0-4 if your only options are 1-3?

SteamReview/SteamReview/Steam.cpp
```diff
@@ -0,0 +1,320 @@
...
+			std::cout << "The category was added succesfully!" << std::endl;
+		}
+		else
+		{
+			std::cout << "something is wrong, the category wasn't be created..." << std::endl;
+		}
+
+		//back to MainMenu
+		std::cout << "Press enter to come back to Main menu" << std::endl;
+		if (std::cin.peek() == '\n')
```
JDAlzate, line 102: Same here as before. Every time you're doing this "Press enter to..." you're neglecting to handle any case in which the user doesn't press enter.

SteamReview/SteamReview/Steam.cpp
```diff
@@ -0,0 +1,320 @@
...
+		std::cin.clear();
+		std::cin.ignore(100, '\n');
+		std::cout << "Invalid Option, please enter a valid Option " << std::endl;
+		std::cin >> InputValue;
+	}
+	return InputValue;
+};
+
+//--------------------------------------------------------------
+FGame FSteam::AddANewGame()
```
JDAlzate, line 193: Prepositions are not very necessary in function and variable names, since the names are clear without them and they just make your code unnecessarily longer. `AddANewGame` and `AddNewGame` are the same thing basically, but the latter is actually easier to read. This goes for this function and `DeleteACategory`

SteamReview/SteamReview/Steam.cpp
```diff
@@ -0,0 +1,320 @@
...
+			OpenDisplayGames();
+			break;
+		case EMenuCommand::Exit:
+			break;
+		}
+	}
+}
+
+//---------------------------------------------------------------
+int FSteam::InputValidate(int InputValue, const int IntLowValueToValid, const int IntHighValueToValid)
```
JDAlzate, line 180: InputValidate is not a very clear function name at first glance. Remember that functions ideally start with a verb, so `ValidateInput` would be better.

JDAlzate, line 180: This method could also be const.

SteamReview/SteamReview/Steam.cpp
```diff
@@ -0,0 +1,320 @@
...
+		std::system("cls");
+		std::cout << "------------------------------------------\nPlease Select the category where you want to add a game" << std::endl;
+		CategoryContainer.PrintCategories();
+
+		int CategoryChosen;
+		std::cin >> CategoryChosen;
+		InputValidate(CategoryChosen, 0, CategoryContainer.GetCurrentCategoryAmount());
+
+		//adding the game to category
+		if (CategoryContainer.AddGameToCategory(CategoryChosen - 1, InGame))
```
JDAlzate, line 250: Careful here. You're allowing the input to be 0, but you're using the input - 1. That means that if you choose 0, you'd be giving an input of -1.

SteamReview/SteamReview/Steam.cpp
```diff
@@ -0,0 +1,320 @@
...
+{
+	//creating a new category
+	std::system("cls");
+	std::cout << "Please enter the name for the new category" << std::endl;
+	std::string CategoryName;
+	std::cin.ignore(500, '\n');
+	std::getline(std::cin, CategoryName);
+	if (CategoryContainer.AddCategory(CategoryName))
+	{
+		return true;
```
JDAlzate, line 282: Remember that this `if (blah) true; else false;` is unnecessarily redundant.
You could simply use `return CategoryContainer.AddCategory(CategoryName);`

Or, if you want to be more explicit (which I generally personally prefer, but it's absolutely not necessary and just something I think is more readable) you could create a bool and return it.
```
const bool bWasAdditionSuccessful = CategoryContainer.AddCategory(CategoryName);
return bWasAdditionSuccessful;
```
Whichever way you prefer is okay :) As long as it's not this redundant if-else

SteamReview/SteamReview/Steam.cpp
```diff
@@ -0,0 +1,320 @@
...
+//------------------------------------------------------------------------
+bool FSteam::DeleteACategory()
+{
+	if (CategoryContainer.IsEmpy())
+	{
+		std::cout << "You aren't saved any category yet..." << std::endl;
+		return false;
+	}
+	
+	else
```
JDAlzate, line 300: This else is unnecessary. You can already clearly see that if the container is empty, you are returning false, so this block of code would never execute if CategoryContainer.IsEmpty is true. 

This is called an early return and it's very useful for readability. With it, you can Remove the else entirely and just leave the code that was inside it.

SteamReview/SteamReview/Steam.cpp
```diff
@@ -0,0 +1,320 @@
...
+		//Deleting a category
+		std::system("cls");
+		CategoryContainer.PrintCategories();
+		std::cout << "-----------------------------------------------" << std::endl;
+
+		std::cout << "Selected the category to delete" << std::endl;
+		int CategorySelected;
+		std::cin >> CategorySelected;
+		InputValidate(CategorySelected, 0, CategoryContainer.GetCurrentCategoryAmount());
+		if (CategoryContainer.DeleteCategory(CategorySelected - 1))
```
JDAlzate, line 311: Here's another example of unnecessarily complicated return statements for a single condition.

SteamReview/SteamReview/Main.cpp
```diff
@@ -0,0 +1,7 @@
+#include "Steam.h"
+
+int main()
+{
+	FSteam Steam = FSteam();
```
JDAlzate, line 5: Remember that the ` = FSteam()` is fine but unnecessary. Simply having `FSteam Steam;` will call the default constructor.

JDAlzate (changes requested):
Nice job! Your code is looking very, very nice! I can clearly see your great improvement compared to previous reviews! Style-wise, there are a few comments in this review.

The only important comment is to test your code to make sure your validations are working! If you had tested with out-of-range values, you would've noticed that your validations are not working. Other than that, great job :)

PR #10, opened 2022-08-17
Steam With References

simonkytr:


SteamReview/SteamReview/Category.h
```diff
@@ -12,14 +12,16 @@ class FCategory
...
+	bool AddGame(const FGame& InGame);
+	bool GetGame(const int GameIndex, FGame& OutGame) const;
 
 	bool IsFull() const;
 	bool IsValid() const;
 
-	std::string GetCategoryName() const;
-	void DisplayCategory() const;
+	const std::string& GetCategoryName() const;
+	void DisplayCategory(FGame& OutGame) const;
```
DvdHalliday, line 26: Now, this is really confusing, why would you have an out parameter in a method named DisplayCategory, if that method is named like that, I expect its behaviour will be restricted to displaying the category, not doing anything at all with the object and variables it deals with. I think that is also why this method is const, 'cause it's not supposed to modify anything, just display. It doesn't really make sense to add an out parameter to a const method, cause you won't even be able to modify it, it should be returned exactly how it was passed to the function.

SteamReview/SteamReview/Category.cpp
```diff
@@ -55,13 +69,22 @@ bool FCategory::IsValid() const
 };
 
 //-------------------------------------------------
-void FCategory::DisplayCategory() const
+void FCategory::DisplayCategory(FGame& OutGame) const
```
DvdHalliday, line 72: You see? you don't even ever use that OutGame

SteamReview/SteamReview/Steam.cpp
```diff
@@ -110,11 +119,24 @@ void FSteam::OpenCategoryManager()
 void FSteam::OpenDisplayGames()
 {
 	std::system("cls");
-	Uncategorized.DisplayCategory();
+	FGame UncategorizedGame;
```
DvdHalliday, line 122: This is such a strange behaviour, and it has to do with the way you changed your DisplayCategory function, you now have to create a Game that is never used, only to be able to call the DisplayCategory method.

SteamReview/SteamReview/Steam.cpp
```diff
@@ -110,11 +119,24 @@ void FSteam::OpenCategoryManager()
...
 	for (int i = 0; i < CategoryContainer.GetCurrentCategoryAmount(); i++)
 	{
 		std::cout << "----------------------------------------------------------------" << std::endl;
-		CategoryContainer.GetCategory(i).DisplayCategory();
+		FCategory Category;
+		
+		if (CategoryContainer.GetCategory(i, Category))
+		{
+			std::cout << "Your Categories" << std::endl;
+			FGame Game;
```
DvdHalliday, line 133: Same here, that Game is completely useless.

SteamReview/SteamReview/Steam.cpp
```diff
@@ -185,24 +205,24 @@ FGame FSteam::AddNewGame()
 	std::cout << "Please enter Day ";
```
DvdHalliday, line 205: This is a bit vague, what day? any day? Mind elaborating a bit, same for the month and year messages

SteamReview/SteamReview/Steam.cpp
```diff
@@ -185,24 +205,24 @@ FGame FSteam::AddNewGame()
...
 
 	std::cout << "Please enter Year ";
 	int InYear;
 	std::cin >> InYear;
-	InYear = ValidateInput(InYear, 1990, 2022);
+	ValidateInput(InYear, 1990, 2022);
 
 	//Creating a Game
-	return FGame(InGameName, InStudioName, InDay, InMonth, InYear);
+	InGame = FGame(InGameName, InStudioName, InDay, InMonth, InYear);
```
DvdHalliday, line 221: This is weird, you're treating this parameter as an Out parameter, so maybe change its name to reflect what you're doing with it.

SteamReview/SteamReview/Steam.cpp
```diff
@@ -213,7 +233,7 @@ bool FSteam::AddNewGameToCategory(const FGame InGame)
```
DvdHalliday, line 233: Comment for line 232: Same as before, 'follow' mightn't be the word you're looking for. Try switching it to continue and you're good to go

SteamReview/SteamReview/Steam.cpp
```diff
@@ -226,19 +246,31 @@ bool FSteam::AddNewGameToCategory(const FGame InGame)
...
+			FCategory Category;
+			if(CategoryContainer.GetCategory(CategoryChosen, Category))
 			{
-				std::cout << "------------------------------------------\n" << std::endl;
-				std::cout << "The game is going to add to the category: " << CategoryContainer.GetCategory(CategoryChosen - 1).GetCategoryName() << std::endl;
-				return true;
+				if (CategoryContainer.AddGameToCategory(CategoryChosen, InGame))
+				{
+					std::cout << "------------------------------------------\n" << std::endl;
+					std::cout << "The game is going to add to the category: " << Category.GetCategoryName() << std::endl;
```
DvdHalliday, line 261: The game isn't adding anything to the category, it is instead going to be added. It has in fact, by this moment, already been added, so 'The game has been to the category' would make much more sense as a message for the user

SteamReview/SteamReview/Steam.cpp
```diff
@@ -226,19 +246,31 @@ bool FSteam::AddNewGameToCategory(const FGame InGame)
...
+				{
+					std::cout << "The game couldn't be added to the category: " << Category.GetCategoryName() << std::endl;
+					return false;
+				}
 			}
 			else
 			{
 				std::cout << "------------------------------------------\n" << std::endl;
-				std::cout << "The game couldn't be added to the category: " << CategoryContainer.GetCategory(CategoryChosen - 1).GetCategoryName() << std::endl;
+				std::cout << "Something is wrong" << std::endl;
```
DvdHalliday, line 273: Something WENT wrong, you might want to add a lil more info into this, if it is already being shown to the user, perhaps be clearer on what it is exactly what went wrong in here, so he could know how to ask for assistance.

SteamReview/SteamReview/Steam.cpp
```diff
@@ -226,19 +246,31 @@ bool FSteam::AddNewGameToCategory(const FGame InGame)
...
 				return false;
 			}
 		}
```
DvdHalliday, line 276: I see here that you actually repeat the behaviour here, so all the comments above actually still apply. Remember it's empty, not empy, and you made another mistake in the comment below, the user hasn't*

SteamReview/SteamReview/Steam.cpp
```diff
@@ -290,8 +322,10 @@ bool FSteam::DeleteACategory()
 		std::cout << "Selected the category to delete" << std::endl;
```
DvdHalliday, line 322: Select instead of selected

SteamReview/SteamReview/Steam.cpp
```diff
@@ -290,8 +322,10 @@ bool FSteam::DeleteACategory()
...
 		int CategorySelected;
 		std::cin >> CategorySelected;
+		ValidateInput(CategorySelected, 1, CategoryContainer.GetCurrentCategoryAmount());
 
-		CategorySelected = ValidateInput(CategorySelected, 1, CategoryContainer.GetCurrentCategoryAmount()) - 1;
+		//Subtract 1 to select the current object inside of the array
+		CategorySelected = CategorySelected - 1;
 		return CategoryContainer.DeleteCategory(CategorySelected);
 	}
 	return false;
```
DvdHalliday, line 331: Oh it all makes sense on why you have that ComeBackMainMenu function, I'm still not fully convinced by the name though, as it is not clear that it will require some interaction with the user

DvdHalliday (commented):
Nice. Couple of comments though, some have to do with typos or grammar mistakes, others really need some attending to. Good luck

SteamReview/SteamReview/Game.h
```diff
@@ -11,11 +11,13 @@ class FGame
 	int Year;
 public:
 	FGame();
-	FGame(const std::string InName, const std::string InStudio, const int InDay, const int InMonth, const int InYear);
+	~FGame();
+	FGame(const FGame& OtherGame);
+	FGame(const std::string& NameRef, const std::string& GameStudioRef, const int& DayRef, const int& MonthRef, const int& YearRef);
```
JDAlzate, line 16: I see you corrected the built-in type references, but you missed some here :)

SteamReview/SteamReview/Game.cpp
```diff
@@ -9,22 +9,38 @@ FGame::FGame()
 };
```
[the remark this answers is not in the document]
JDAlzate: I see you added a commit with changes for David's comments. But some of them have not been addressed (like this one). If you don't agree with something or don't understand something, it's good to leave a comment about it, rather than ignoring it.

Or in the case that you guys discussed this in person and decided to not apply the comment, it's still good to leave the conclusion written in the review.

However in this case, I do agree with David :) It increases readability and consistency of whitespace.

SteamReview/SteamReview/Game.cpp
```diff
@@ -9,22 +9,38 @@ FGame::FGame()
...
+const std::string& FGame::GetGameName() const
 {
 	return NameGame;
 };
 //-----------------------------------------------------
-std::string FGame::GetGameStudio() const
+const std::string& FGame::GetGameStudio() const
 {
 	return GameStudio;
 };
```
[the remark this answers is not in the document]
JDAlzate: Nice! Didn't catch that in the previous review :)

SteamReview/SteamReview/Game.cpp
```diff
@@ -9,34 +9,50 @@ FGame::FGame()
...
+std::string FGame::GetDateReleaseFormat() const
 {
 	return std::to_string(Year) + "/" + std::to_string(Month) + "/" + std::to_string(Day);
 };
 
 //-----------------------------------------------------
 bool FGame::IsValid() const
 {
-	return NameGame.empty() && GameStudio.empty() && Day <= 0 && Month <= 0 && Year <= 0;
+	return !NameGame.empty() && !GameStudio.empty() && Day <= 0 && Month <= 0 && Month > 12 && Year <= 0;
```
JDAlzate, line 57: I see you fixed the empty names issues, but the date is still being validated incorrectly.

SteamReview/SteamReview/CategoryContainer.cpp
```diff
@@ -18,6 +18,19 @@ bool FCategoryContainer::AddCategory(const FCategory InCategory)
...
+FCategoryContainer::~FCategoryContainer()
+{
+	//Destructor
+};
+
+//--------------------------------------------------------
+FCategoryContainer::FCategoryContainer(FCategoryContainer& OtherContainer)
+{
+	CategoryCount = OtherContainer.CategoryCount;
+	memcpy_s(Categories, sizeof(Categories), OtherContainer.Categories, sizeof(OtherContainer.CategoryCount));
```
JDAlzate, line 31: This line has an important bug in it. Try to see if you find it.

SteamReview/SteamReview/CategoryContainer.cpp
```diff
@@ -34,13 +47,18 @@ bool FCategoryContainer::DeleteCategory(const int InCategoryIndex)
 };
```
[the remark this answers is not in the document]
JDAlzate: Nice :) You're catching stuff that a few others of us haven't caught @DvdHalliday

SteamReview/SteamReview/Steam.cpp
```diff
@@ -8,6 +8,14 @@ FSteam::FSteam()
 	ActiveMenuCommand = EMenuCommand::MainMenu;
 };
 
+//---------------------------------------------------------
+FSteam::FSteam(FSteam& OtherSteam)
```
JDAlzate, line 12: Remember that you want to use _const_ references if the original value isn't supposed to change.

SteamReview/SteamReview/Steam.cpp
```diff
@@ -46,7 +53,8 @@ void FSteam::OpenMainMenu()
 void FSteam::OpenAddGame()
 {
 	//adding a new game
-	FGame Game = AddNewGame();
+	FGame Game;
+	AddNewGame(Game);
```
JDAlzate, line 57: Maybe you'd like to handle the scenario of the result of this function being false?

SteamReview/SteamReview/Steam.cpp
```diff
@@ -110,11 +119,24 @@ void FSteam::OpenCategoryManager()
 void FSteam::OpenDisplayGames()
```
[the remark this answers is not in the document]
JDAlzate: Here's another important comment that you didn't address.

SteamReview/SteamReview/Steam.cpp
```diff
@@ -148,21 +167,20 @@ void FSteam::AppRun()
 }
 
 //---------------------------------------------------------------
-int FSteam::ValidateInput(int InputValue, const int LowBoundValue, const int HighBoundValue)
+void FSteam::ValidateInput(int& InputValue, const int LowBoundValue, const int HighBoundValue)
 {
-
+	std::cin >> InputValue;
```
JDAlzate, line 172: Since you're assigning values to InputValue completely inside this function, maybe it would be better to return the input as a return type, rather than as an out parameter.

Ouput parameters are useful in some scenarios, like for example when you want to return a value but also return whether or not the returned value is valid, so you use a bool return type and an out parameter for the other type you want to return.

However, they are not the standard way of returning stuff, that's what the return type is for. With this in mind, although this method works fine, I believe that it would be clearer if you used the return type, the way you were doing before.

Ouput parameters should be special cases, rather than the rule. Especially with built-in types.

SteamReview/SteamReview/Steam.cpp
```diff
@@ -148,21 +167,20 @@ void FSteam::AppRun()
...
 		std::cin.ignore(100, '\n');
 		std::cout << "Invalid Option, please enter a valid Option " << std::endl;
 		std::cin >> InputValue;
 	}
-	return InputValue;
 };
 
 //--------------------------------------------------------------
-FGame FSteam::AddNewGame()
+void FSteam::AddNewGame(FGame& InGame)
```
JDAlzate, line 183: Doesn't seem like you'd want to change this game at all, would you?

In that case, this would be a good place to use a _const_ reference

SteamReview/SteamReview/Steam.cpp
JDAlzate: Yup! I added the previous comment thinking it was indeed an In parameter, but it's not. So maybe it should be named OutGame. 

Maybe it would be good to rename the function as well.

`AddNewGame` implies that the game is being added somewhere, but it's not. Instead, you are creating a new game, so why not call this `CreateNewGame`?

SteamReview/SteamReview/Steam.h
```diff
@@ -18,16 +18,17 @@ class FSteam
...
 	void OpenDisplayGames();
 
-	int ValidateInput(int InputValue, const int LowBoundValue, const int HighBoundValue);
+	void ValidateInput(int& InputValue, const int LowBoundValue, const int HighBoundValue);
 
-	FGame AddNewGame();
-	bool AddNewGameToCategory(const FGame InGame);
+	void AddNewGame(FGame& InGame);
+	bool AddNewGameToCategory(const FGame& InGame);
 	bool AddNewCategory();
```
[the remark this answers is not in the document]
JDAlzate: Agree with this comment, but it wasn't addressed.

SteamReview/SteamReview/Steam.h
```diff
@@ -18,16 +18,17 @@ class FSteam
...
-	bool DeleteACategory();
+	bool DeleteCategory();
 
 	void ComeBackMainManu();
 
 public:
 	FSteam();
-	void AppRun();
+	FSteam(FSteam& OtherSteam);
```
JDAlzate, line 32: Remember that this should be a const reference

JDAlzate (changes requested):
Good job :) But be very careful with reviews. It's great that you're addressing comments before the final submission, but in reviews, you should NEVER ignore comments and you've ignored quite a bit of them.

PR #11, opened 2022-08-24
Pointer_Excercises

simonkytr:
excercise1
Ans: The Results make sense to me because each variable is allocated to different addresses.
pointers1

excercise2
pointers2

exercise3
Ans: The results make sense to me because the variables are allocated in different places to pointers.
pointers3

exercise
Ans: it makes sense because the dereferencing is access to the variable value, furthermore it can be able change this value.
pointers4

exercise5
ans: The results make sense to me because when you add more elements to the elements already initialized to the array pointer, it is going to fill these spaces with random numbers o waste numbers.
pointers5

exercise6
pointers6

### DIFF
--- a/Pointers_Excercises/Excercise1/Source.cpp
+++ b/Pointers_Excercises/Excercise1/Source.cpp
@@ -1,0 +1,25 @@
+#include<iostream>
+
+int main()
+{
+	std::cout << "//////////////////////////////////////////" << std::endl;
+	std::cout << "insert an integer value" << std::endl;
+	int InputValue;
+	std::cin >> InputValue;
+	std::cout << "The value of your variable is: " << InputValue << std::endl;
+	std::cout << "The memory address where your variable is stored is: " << &InputValue << std::endl;
+
+	std::cout << "//////////////////////////////////////////" << std::endl;
+	std::cout << "insert an integer value" << std::endl;
+	int InputValue2;
+	std::cin >> InputValue2;
+	std::cout << "The value of your variable is: " << InputValue2 << std::endl;
+	std::cout << "The memory address where your variable is stored is: " << &InputValue2 << std::endl;
+
+	std::cout << "//////////////////////////////////////////" << std::endl;
+	std::cout << "insert an integer value" << std::endl;
+	int InputValue3;
+	std::cin >> InputValue3;
+	std::cout << "The value of your variable is: " << InputValue3 << std::endl;
+	std::cout << "The memory address where your variable is stored is: " << &InputValue3 << std::endl;
+}
--- a/Pointers_Excercises/Excercise1/Source.cpp
+++ b/Pointers_Excercises/Excercise1/Source.cpp
@@ -1,3 +1,12 @@
+/*Exercises
+Exercise 1 - Understanding Memory Addresses
+Create a program that will ask the user to enter an integer and assign the given value to a
+variable. Then, print the value to the console, and print the memory address of that variable.
+Then, ask for another value and assign that value to the same variable. Print the value and the
+memory address to the console.
+Finally, ask for one more value. This time, assign the value to a different variable. Print the
+value and the memory address to the console. Do the results make sense to you?*/
+
 #include<iostream>
 
 int main()
--- a/Pointers_Excercises/Excercise1/Source.cpp
+++ b/Pointers_Excercises/Excercise1/Source.cpp
@@ -28,7 +28,8 @@ int main()
 	std::cout << "//////////////////////////////////////////" << std::endl;
 	std::cout << "insert an integer value" << std::endl;
 	
-	std::cin >> InputValue;
-	std::cout << "The value of your variable is: " << InputValue << std::endl;
-	std::cout << "The memory address where your variable is stored is: " << &InputValue << std::endl;
+	int InputValue2;
+	std::cin >> InputValue2;
+	std::cout << "The value of your variable is: " << InputValue2 << std::endl;
+	std::cout << "The memory address where your variable is stored is: " << &InputValue2 << std::endl;
 }
--- a/Pointers_Excercises/Excercise1/Source.cpp
+++ b/Pointers_Excercises/Excercise1/Source.cpp
@@ -20,15 +20,15 @@ int main()
 
 	std::cout << "//////////////////////////////////////////" << std::endl;
 	std::cout << "insert an integer value" << std::endl;
-	int InputValue2;
-	std::cin >> InputValue2;
-	std::cout << "The value of your variable is: " << InputValue2 << std::endl;
-	std::cout << "The memory address where your variable is stored is: " << &InputValue2 << std::endl;
+	
+	std::cin >> InputValue;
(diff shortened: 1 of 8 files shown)


Pointers_Excercises/Excercise1/Source.cpp
```diff
@@ -0,0 +1,34 @@
...
+	std::cout << "//////////////////////////////////////////" << std::endl;
+	std::cout << "insert an integer value" << std::endl;
+	int InputValue;
+	std::cin >> InputValue;
+	std::cout << "The value of your variable is: " << InputValue << std::endl;
+	std::cout << "The memory address where your variable is stored is: " << &InputValue << std::endl;
+
+	std::cout << "//////////////////////////////////////////" << std::endl;
+	std::cout << "insert an integer value" << std::endl;
+	int InputValue2;
```
DvdHalliday, line 31: This is not what the exercise is asking for, you're supposed to assign that new value into the same variable, not create a new one.

Pointers_Excercises/Excercise2/Source.cpp
```diff
@@ -0,0 +1,25 @@
...
+	std::cin >> InputValue;
+	std::cout << "The value of variable a is now: " << InputValue << std::endl;
+
+	int* InputValuePtr = &InputValue;
+
+	std::cout << "\nPlease enter a value for variable a: ";
+	std::cin >> *InputValuePtr;
+	std::cout << "The value of variable a is now: " << InputValue << std::endl;
+
+	std::cout << "\nFun fact: I didn’t have to touch a directly to change its value :)" << std::endl;
```
DvdHalliday: That ’ character is being printed as an odd character, not the one you want.

Pointers_Excercises/Excercise3/Source.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+	std::cout << "Please enter a value for variable b: ";
+	int InputValueB;
+	std::cin >> InputValueB;
+	
+
+	std::cout << std::endl;
+
+	std::cout << "Using pointer variable p, I've determined that: " << std::endl;
+	std::cout << "The value of a is " << InputValueA << std::endl;
```
DvdHalliday, line 30: I think you're supposed to do all of this using strictly the p variable. The same applies for the next line and when the code gets repeated.

Pointers_Excercises/Excercise4/Source.cpp
```diff
@@ -0,0 +1,22 @@
...
+int main()
+{
+	int Value = 15;
+	int* ValuePtrA = &Value;
+	int* ValuePtrB = &Value;
+
+	std::cout << "The value of my variable is " << *ValuePtrA << std::endl;
+
+	*ValuePtrA = 6;
+	*ValuePtrB = 2;
```
DvdHalliday: This is again not exactly what the exercise is asking for, "add one" here means that you should add one unit to the current value that's stored in that pointer, so you're essentially being asked to do *ValuePtrA += 1.

Pointers_Excercises/Excercise6/Source.cpp
```diff
@@ -0,0 +1,38 @@
...
+	int* ArrayPtr = Array;
+
+	std::cout << "I've looped through the following numbers :" << std::endl;
+	for (int i = 0; i < 10; i++)
+	{
+		ArrayPtr[i] = rand() % 100;
+		std::cout << ArrayPtr[i] << " ";
+	}
+
+	std::cout << "\nI've looped through the following numbers (BACKWARDS) :" << std::endl;
```
DvdHalliday: The exercise isn't really asking you to print it backwards, it's nice it came out alright tho.

Pointers_Excercises/Excercise6/Source.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <stdlib.h>     /* srand, rand */ 
+
+int main()
+{
+
+	int Array[10];
+	int* ArrayPtr = Array;
+
+	std::cout << "I've looped through the following numbers :" << std::endl;
+	for (int i = 0; i < 10; i++)
```
DvdHalliday: Here you're not actually proceeding as you were asked to do. You should not initialize an indexer integer. The exercise asks of you to loop through the array strictly using the pointer which you can manipulate using the pointer arithmetic.

Pointers_Excercises/Excercise6/Source.cpp
```diff
@@ -0,0 +1,38 @@
...
+	{
+		ArrayPtr[i] = rand() % 100;
+		std::cout << ArrayPtr[i] << " ";
+	}
+
+	std::cout << "\nI've looped through the following numbers (BACKWARDS) :" << std::endl;
+	int Sum = 0;
+	for (int i = 10; i > 0; i--)
```
DvdHalliday: Same here, you're supposed to iterate using only the pointer ArrayPtr

Pointers_Excercises/Excercise7/Source.cpp
```diff
@@ -0,0 +1,38 @@
...
+Depending on whether you validated the pointer or not, your program may have crashed. This is
+one of the dangers of pointers.
+Accessing null pointers will crash your application, so it’s important to be sure to add validity
+checks when there is even the slightest possibility that the value of a pointer will be null!*/
+
+#include <iostream>
+
+void IncrementNumber(int* NumberPtr)
+{
+	if (NumberPtr != nullptr)
```
DvdHalliday, line 19: This is not wrong, however when dealing with pointers i think it's safe to say we can now use just the bool checker for them, so instead of the NumberPtr != nullptr, it's well known that this can also be checked by doing simply if (NumberPtr)

Pointers_Excercises/Excercise7/Source.cpp
```diff
@@ -0,0 +1,38 @@
...
+	{
+		(*NumberPtr)++;
+	}
+	else
+	{
+		std::cout << "That's not possible!!!";
```
DvdHalliday, line 25: Oh it's nice you added some else behaviour

DvdHalliday (changes requested):
It's nice altogether. Perhaps be a lil' more careful when reading the exercises, or even if something isn't clear about the statements feel free to ask a peer if an exercise is in fact about what you thought it was.

Also, why is there only one vcxproj file? shouldn't there be one for each project?

Pointers_Excercises/Excercise3/Source.cpp
JDAlzate: Yep, this is not doing exactly what is asked in the exercise. You shouldn't be using InputValueA or InputValueB

Pointers_Excercises/Excercise7/Source.cpp
JDAlzate: Yep. You can do it both ways. It's more common to see the `if (NumberPtr)` alone, but however you prefer to use it is fine.

JDAlzate (changes requested):
Just one comment there :)

PR #12, opened 2022-08-26
Memory_Excercises

simonkytr:
Excercise1
memory1
Excercise2
memory2
Excercise3
memory3
Excercise4
Memory5
Excercise7
Memory7
Excersice 11
Memory11


Memory_Excercises/Excercise2/Source.cpp
```diff
@@ -0,0 +1,56 @@
...
+	std::cout << HeapInt1 << std::endl;
+	std::cout << HeapInt2 << std::endl;
+	std::cout << HeapInt3 << std::endl;
+
+	delete HeapInt1;
+	delete HeapInt2;
+	delete HeapInt3;
+
+	std::cout << "What can you notice from the memory addresses in the stack vs the memory addresses in the heap ?" << std::endl;
+	std::cout << "I can notice that the heap memory address has a more enormous Address value than the stack address value" << std::endl;
```
DvdHalliday, line 54: Hmmm I think the exercise here is supposed to ask about what exactly can you notice in the way that new variables are being stored according to the type of memory they were set to. For instance, you might want to compare the values of the memory between one another of the stack memory-stored values, then do the same but for the heap-stored ones. A wee hint is that your stack variables are exactly 20Hex bits apart from one another, why is this? Does this also happen for the values stored on the heap memory?

Memory_Excercises/Excercise3/Source.cpp
```diff
@@ -0,0 +1,46 @@
...
+function. You�ll find that there is no way to do it - you�ve created two memory leaks. These leaks
+will live in memory for as long as the program is running and you won�t be able to store any new
+data in those memory locations.
+Print to the console why each of those functions creates a memory leak.*/
+
+#include <iostream>
+
+void MemoryLeak1()
+{
+	int* IntPtr = new int;
```
DvdHalliday, line 28: I'm not really sure here, I'd appreciate JD's comment, but since you're initialising a built-in type here, but you're not assigning any value to it, you might as well be leaving it set to whatever value it used to be before, since this is heap memory I think it could actually still be 'dd' for the four bytes of memory an int takes. So, I'm not quite certain here but that might cause some trouble, it could... perhaps still be recognized as free memory for the heap memory to store new values into?

Memory_Excercises/Excercise3/Source.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+//---------------------------
+int main()
+{
+	MemoryLeak1();
+
+	int* Pointer = nullptr;
+	MemoryLeak2(Pointer);
+	std::cout << "why each of those functions creates a memory leak?" << std::endl;
+	std::cout << "each function creates a memory leak because when we reserved some heap memory and we don't using\nfurthermore we don't free this memory simply then we had reserved memory that is beeing useless";
```
DvdHalliday, line 45: Feel free to add the messages in different lines or std::couts, to help with readability

DvdHalliday: Also, I think you're missing a lot on the explanation for the second memory leak, the exercise is also asking you to explain why, even if we wanted to, we couldn't do a delete on that Pointer back in the main function. If you check the memory viewer, you'll soon realise that something quite strange is happening with the addresses.

Memory_Excercises/Excercise4/Source.cpp
```diff
@@ -0,0 +1,30 @@
...
+	
+	if (Pointer != nullptr)
+	{
+		std::cout << "The pointer is valid" << *Pointer << std::endl;
+	}
+	else
+	{
+		std::cout << "The pointe is invalid" << std::endl;
+	}
+	std::cout << "The pointer is working but the address that it was pointing to has been deleted";
```
DvdHalliday, line 29: I feel like this explanation could be more elaborate (?)

Memory_Excercises/Excercise5/ClassA.h
```diff
@@ -0,0 +1 @@
+
```
DvdHalliday, line 1: Careful with this kind of empty files, they shouldn't be added to any commit

Memory_Excercises/Excercise5/Source.cpp
```diff
@@ -0,0 +1,126 @@
+/*Exercise 5 - Dynamic Allocation for Array Sizes
+*Create a Student class that contains a char array of size 5 for a name and an int for an age. Ask
```
DvdHalliday, line 2: o;; comment block according to standards ✓✓✓

DvdHalliday: although you're missing a space after every asterisk lol

Memory_Excercises/Excercise5/Source.cpp
```diff
@@ -0,0 +1,126 @@
...
+	int StudentAge;
+
+	FStudent()
+	{
+		NameStudent = nullptr;
+		StudentAge = 0;
+	}
+
+	//----------------------------------------------
+	FStudent( int NewAge)
```
DvdHalliday, line 28: Unnecesary space after the (

Memory_Excercises/Excercise5/Source.cpp
```diff
@@ -0,0 +1,126 @@
...
+	{
+		NameStudent = nullptr;
+		StudentAge = NewAge;
+	}
+
+	//----------------------------------------------
+	~FStudent()
+	{
+		delete NameStudent;
```
DvdHalliday, line 37: Be careful here, you're supposed to set the NameStudent variable to nullptr after you delete it.

Memory_Excercises/Excercise5/Source.cpp
```diff
@@ -0,0 +1,126 @@
...
+	//----------------------------------------------
+	void SetName(const char* NewName)
+	{
+		/*Exercise 9: At this point, it�s obvious that keeping track of all memory allocations in a program can be a hassle. In the future, 
+		*we�ll discuss techniques that will make life a lot easier for us.
+		*/
+
+		if (NameStudent != nullptr)
+		{
+			NameStudent = nullptr;
```
DvdHalliday, line 49: MMmmm, you're actually not fixing the memory leak this way, you should also delete what's saved into this variable before setting it to nullptr

Memory_Excercises/Excercise5/Source.cpp
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		int NameLenght = 0;
+		for (int i = 0; i < 32; i++)
+		{
+			if (NewName[i] != '\0')
+			{
+				NameLenght++;
+			}
```
DvdHalliday, line 58: perhaps add a break as an else statement (?)

Memory_Excercises/Excercise5/Source.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+		NameStudent = new char[NameLenght + 1];
+
+		memcpy_s(NameStudent, sizeof(char) * NameLenght + 1, NewName, sizeof(char) *
+			NameLenght + 1);
+	};
+};
+
+//--------------------------------------------------------------------
+int ValidateAge()
```
DvdHalliday, line 69: o; input validation, lovely!

Memory_Excercises/Excercise5/Source.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+	for (int i = 0; i < StudentsAmount; i++)
+	{
+		std::cout << "student " << i + 1 << std::endl;
+		std::cout << "Name " << Students[i].NameStudent << std::endl;
+		std::cout << "Age " << Students[i].StudentAge << std::endl;
+	}
+
+	std::cout << "Excercise 7) why deleting the student object does not delete the name�s char array?" << std::endl;
+	std::cout << "the name�s char array doesn't delete because when the class is deleted, just is deleting the memory of the class but not the variable that is allocated in the memory stack" << std::endl;
```
DvdHalliday, line 124: Nice, this is a great answer.

Memory_Excercises/Excersice10/Source.cpp
```diff
@@ -0,0 +1,12 @@
+#include "ClassA.h"
+#include "ClassB.h"
+#include <iostream>
+
+int main()
+{
+	FClassA ClassA;
+	FClassB ClassB;
+
+	std::cout << "ans 10: the program doesn't work because the Main is calling the header file of both classes so it's making a cycle" << std::endl;
```
DvdHalliday, line 10: Perhaps elaborate a bit more... Think of what the exercise tells you to consider, (#include stuff)

Memory_Excercises/Excersice10/Source.cpp
```diff
@@ -0,0 +1,12 @@
...
+	std::cout << "ans 11:The Forwards declarations are so useful when we need to reduce the debug's time furthermore whether we want to is so useful when we need to declare as a pointer";
```
DvdHalliday, line 11: How are they actually helping to reduce debug time?

DvdHalliday (changes requested):
Great job.

I think the pic in the pull request post for the fourth exercise is a bit misleading, it's a screenshot of the fifth exercise or one of the exercise that also use this same project, however it's not clear if it's the fifth exercise, and even if it is, it would be odd that you can add more than four characters to a name and they get stored correctly.

Memory_Excercises/Excercise2/Source.cpp
JDAlzate, line 54: I'm actually not sure how accurate that statement is, but what I wanted you to notice is that the values in the stack are a lot closer together than the values in the heap :)

Memory_Excercises/Excercise3/Source.cpp
JDAlzate: I'd understand that the 'dd' you're refering to is the value that can be seen in the memory view, right? In that case, it's not a problem. The reason it's value is something like dd is the optimizations that are done by running the program in the Debug configuration, so it provides guards for uninitialized memory. However, that's not the way that the program knows which memory is free and which isn't, because those guards are not used in Release mode, for example. 

We can discuss it further on Monday with both of you if you want, but that particular issue you're describing wouldn't occur :) I don't believe so @DvdHalliday

Memory_Excercises/Excercise3/Source.cpp
JDAlzate: Yep, I was looking for a bit more there. It's not only that it's a leak because we are not freeing it, but rather because at any point we simply are unable to free it anymore.

Memory_Excercises/Excercise4/Source.cpp
JDAlzate: x2

Memory_Excercises/Excercise5/Source.cpp
JDAlzate: Yep! This is a memory leak! Each call to SetName will create new memory without freeing the old one

Memory_Excercises/Excercise5/Source.cpp
JDAlzate: Yep, better to break when there's no more need to iterate. Also, what if the name was longer than 32 characters?

The conditions in your for loop don't always need to be `i < x`. They can be whatever you want. So you can write something like 
```
for (int i = 0; NewName[i] != 0; i++)
{
    NameLength++;
}
```

That way you don't depend on some arbitrary max length
That way you don't depend on some arbitrary length

Memory_Excercises/Excercise5/Source.cpp
JDAlzate: Careful! It's the char array in the heap that's not being deleted, it has nothing to do with a variable in the stack.

Memory_Excercises/Excersice10/ClassB.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+class FClassB;
+class FClassA
+{
+	FClassB ClassB;
```
JDAlzate, line 6: Shouldn't this be a pointer?

Memory_Excercises/Excersice10/Source.cpp
JDAlzate: x2, doesn't seem like forward declarations are very clear here hehe

JDAlzate (changes requested):
Some of the answers to the questions are very vague, seems to me like you didn't fully understand it and decided to write what you could instead of asking 👀

PR #13, opened 2022-09-09
Array_ vector Exercise

simonkytr:


Array_Vector_Exercises/Exercise1/shape.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+ float FShape::GetArea() const
+{
+	 return 0;
+}
+
+float FShape::GetPerimeter() const
+{
+	return 0;
+}
```
Sergio6040, line 27: I think there is no need to make an .cpp file for this abstract class, for example I only create the .h file and declate the virtual funtions as: 

```
virtual float GetArea() const = 0;
```

Array_Vector_Exercises/Exercise1/Square.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+#include "shape.h"
+
+class FSquare : public FShape
+{
+private:
+	float Lenght;
+	float Height;
```
Sergio6040, line 8: 😅
As it is a square both sides have the same length, otherwise it would be a rectangle xd

Array_Vector_Exercises/Exercise1/square.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+float FSquare::GetArea() const
+{
+	return Lenght * Height;
+}
+
+float FSquare::GetPerimeter() const
+{
+	return (Lenght * 2) + (Height * 2);
+}
```
Sergio6040, line 30: nice 👌

Array_Vector_Exercises/Exercise1/Menu.cpp
```diff
@@ -0,0 +1,379 @@
...
+		std::cout << "Invalid Option, please enter a valid value " << std::endl;
+		std::cin >> UserInput;
+	};
+	return UserInput;
+}
+
+//-----------------------------------------------------------------------------
+void FMenu::ComeBackMainManu()
+{
+	while (std::cin)
```
Sergio6040: wuutt?? xd this is interesting, what are you doing here?

Array_Vector_Exercises/Exercise1/Menu.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+	while (std::cin.fail())
+	{
+		std::cin.clear();
+		std::cin.ignore(100, '\n');
+		std::cout << "Invalid Option, please enter a valid value " << std::endl;
+		std::cin >> UserInput;
+	};
+	return UserInput;
+}
```
Sergio6040, line 362: here maybe you can use an template, don't you think so?

Array_Vector_Exercises/Exercise1/Menu.cpp
```diff
@@ -0,0 +1,379 @@
...
+			AskForBuildInArray();
+			break;
+		case EMenuCommand::StaticArray:
+			AskForStaticArray();
+			break;
+		case EMenuCommand::DynamicArray:
+			AskForDynamicArray();
+			break;
+		case EMenuCommand::Exit:
+			break;
```
Sergio6040, line 314: This case is not necessary

Array_Vector_Exercises/Exercise1/Menu.cpp
```diff
@@ -0,0 +1,379 @@
...
+				delete Shapes[i];
+			}
+			ActiveMenuCommand = EMenuCommand::MainMenu;
+		}
+
+	} while (UserInput != 3);
+}
+
+//---------------------------------------------------------------------------------
+bool FMenu::CreateShape(FShape*& Shape)
```
Sergio6040: ufff a pointer reference parameter, maybe it would be easier just to return an FShape, no??

Array_Vector_Exercises/Exercise1/Menu.cpp
```diff
@@ -0,0 +1,379 @@
...
+		PrintArray(*Shapes[i]);
+	}
+
+	//freeing Memory
+	for (int i = 0; i < SizeArray; i++)
+	{
+		delete Shapes[i];
+	}
+	delete[] Shapes;
+
```
Sergio6040, line 91: I know that are just 3 lines of code, but you could make this a function to reuse it in the other methods, I think you can acces the array by passing it as an pointer parameter. :p

Array_Vector_Exercises/Exercise1/Menu.cpp
```diff
@@ -0,0 +1,379 @@
...
+	for (int i = 0; i < 5; i++)
+	{
+		std::cout << "--------------------------------- " << std::endl;
+		std::cout << "Shape #" << i + 1 << std::endl;
+
+		std::cout << "--------------------------------- " << std::endl;
+		std::cout << "Shape #" << i + 1 << std::endl;
+		PrintArray(*Shapes[i]);
+	}
```
Sergio6040, line 128: This is also repeated.

Array_Vector_Exercises/Exercise1/TStaticArray.h
```diff
@@ -0,0 +1,98 @@
...
+
+	//---------------------------------------------------------
+	bool IsEmpty() const
+	{
+		if (SizeArray == 0)
+		{
+			return true;
+		}
+		return false;
+	};
```
Sergio6040: here you can use
```
return SizeArray == 0;
```
 instead.

Array_Vector_Exercises/Exercise1/TStaticArray.h
```diff
@@ -0,0 +1,98 @@
...
+			ArrayContainer[i] = StaticArray[i];
+		}
+		for (int i = 0; i < SizeArray; i++)
+		{
+			StaticArray[i] = SwapArray[i];
+		}
+		for (int i = 0; i < SizeArray; i++)
+		{
+			SwapArray[i] = ArrayContainer[i];
+		}
```
Sergio6040, line 92: great way, but I think you can do this in the same for loop

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+	{
+		Append(DynamicArray);
+	}
+
+	TDynamicArray& operator= (const TDynamicArray<T>& OtherDynamicArray)
+	{
+		Clear();
+		Append(OtherDynamicArray);
+		return *this;
+	}
```
Sergio6040, line 34: well done 👌

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+
+			for (int i = 0; i < SizeArray; i++)
+			{
+				ContainerArray[i] = DynamicArray[i];
+			}
+			delete[] DynamicArray;
+
+			DynamicArray = ContainerArray;
+		}
+	}
```
Sergio6040, line 131: in this case it would also be necessary to delete the ContainerArray, right?

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+
+	//--------------------------------------------------------------------------
+	void Clear()
+	{
+		SizeArray = 0;
+	}
```
Sergio6040, line 137: remember to also clear the array

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+		}
+		else
+		{
+			DynamicArray[SizeArray] = InType;
+			SizeArray++;
+		}
+	}
+
+	//-----------------------------------------------------------------------------
+	void Pushpop()
```
Sergio6040: Do you mean PopBack()?? xd

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+			{
+				Reserve(CapacityArray * 2);
+			}
+				
+			for (int i = 0; i < SizeArray; i++)
+			{
+				DynamicArray[Index + 1] = DynamicArray[Index];
+			}
+			DynamicArray[Index] = InType;
+			SizeArray++;
```
Sergio6040: this isn't working as is intended, you are not moving all the values +1 position, and also you are losing one value

image

I made this test :p, the resoult should be 71234567890

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+			DynamicArray[Index] = DynamicArray[Index + 1];
+		}
+		SizeArray--;
+	}
+
+	//----------------------------------------------------------------------------
+	void Resize(const int NewSize)
+	{
+		SizeArray = NewSize;
+	}
```
Sergio6040, line 197: here you need to call, the reserve función to avoid errors

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+		{
+			Array1.Reserve(Index);
+			Array1.PushBack(DynamicArray[i]);
+		}
+		for (int i = Index; i < CapacityArray; i++)
+		{
+			Array1.Reserve(CapacityArray - Index);
+			Array2.PushBack(DynamicArray[i]);
+		}
+	}
```
Sergio6040, line 241: noice, great methods 👏

Sergio6040 (changes requested):
Great Work, there are some things to fix 👍

Array_Vector_Exercises/Exercise1/shape.cpp
simonkytr: hmm. thanks, I've never thought before

Array_Vector_Exercises/Exercise1/Square.h
simonkytr: XD, To give it the power

Array_Vector_Exercises/Exercise1/Menu.cpp
simonkytr: what, what, I didn't notice before

Array_Vector_Exercises/Exercise1/Menu.cpp
simonkytr: yep, but JD tell me its unnesesary

Array_Vector_Exercises/Exercise1/Menu.cpp
simonkytr: jajajaj I didnt notice before

Array_Vector_Exercises/Exercise1/Menu.cpp
simonkytr: thaks

Array_Vector_Exercises/Exercise1/Menu.cpp
simonkytr: thats true, thanks

Array_Vector_Exercises/Exercise1/TDinamicArray.h
simonkytr: it's no necessary

Array_Vector_Exercises/Exercise1/TDinamicArray.h
simonkytr: JD explained to us that it must not be deleted

Array_Vector_Exercises/Exercise1/shape.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
```
JDAlzate, line 1: Careful with the file name, you're not following pascal case

Array_Vector_Exercises/Exercise1/shape.h
```diff
@@ -0,0 +1,16 @@
...
+#include <iostream>
+
+class FShape
+{
+protected:
+	std::string KindOfShape;
+public:
+	FShape();
+	
+	virtual std::string GetKindOfShape() const;
```
JDAlzate, line 11: Why make this virtual? Do you ever want it to _not_ return KindOfShape?

Array_Vector_Exercises/Exercise1/Square.h
JDAlzate: Yep, this needs to change.

Array_Vector_Exercises/Exercise1/Menu.cpp
```diff
@@ -0,0 +1,380 @@
...
+					std::cout << "Shape #" << i + 1 << std::endl;
+					PrintArray(*Shapes[i]);
+				}
+			}
+
+			std::cout << "Press 1... to back previus menu" << std::endl;
+			int MenuInput;
+			std::cin >> MenuInput;
+
+			while (std::cin.fail() || MenuInput != 1)
```
JDAlzate, line 224: Isn't this the logic of your validate input function? Why not use it here.

Array_Vector_Exercises/Exercise1/Menu.cpp
```diff
@@ -0,0 +1,380 @@
...
+		std::cout << "Insert square's heignt: ";
+		float InHeight;
+		InHeight = ValidateInputFloat(InHeight);
+
+		return new FSquare(InLenght, InHeight);
+	}
+}
+
+//---------------------------------------------------------------------------------
+void FMenu::PrintArray(const FShape& Shape) const
```
JDAlzate, line 285: Why is this called print array if it prints a single shape?

JDAlzate: Also, this works correctly, but since we're manipulating FShaps with pointers, it may be simpler to use a shape pointer as a parameter. What do you think?

Array_Vector_Exercises/Exercise1/Menu.cpp
```diff
@@ -0,0 +1,380 @@
...
+			AskForDynamicArray();
+			break;
+		case EMenuCommand::Exit:
+			break;
+		}
+	}
+}
+
+//------------------------------------------------------------------------------
+int FMenu::ValidateInput(int& UserInput, const int LowerBounce, const int HigherBounce)
```
JDAlzate, line 320: We mentioned this in earlier reviews. Why would you want to have an out parameter AND a return type that both return the same thing? Why not just leave the return type, what's the purpose of the UserInput variable?

JDAlzate: Also, Bounce is misspelled here, it should be Bound.

Array_Vector_Exercises/Exercise1/Menu.cpp
```diff
@@ -0,0 +1,380 @@
...
+		std::cin.clear();
+		std::cin.ignore(100, '\n');
+		std::cout << "Invalid Option, please enter a valid Option " << std::endl;
+		std::cin >> UserInput;
+	};
+	return UserInput;
+}
+
+//---------------------------------------------------------------------------------
+int FMenu::ValidateInput(int& UserInput)
```
JDAlzate, line 335: Same here as the previous comment. Also, remember that since this isn't really validating an input, but returning a valid input, this may be more appropriately named. Something like GetValidInput or something like that.

Array_Vector_Exercises/Exercise1/Menu.cpp
```diff
@@ -0,0 +1,380 @@
...
+		std::cin.clear();
+		std::cin.ignore(100, '\n');
+		std::cout << "Invalid Option, please enter a valid Option " << std::endl;
+		std::cin >> UserInput;
+	};
+	return UserInput;
+}
+
+//------------------------------------------------------------------------------
+float FMenu::ValidateInputFloat(float& UserInput)
```
JDAlzate, line 350: Same here x3

Array_Vector_Exercises/Exercise1/Menu.cpp
```diff
@@ -0,0 +1,380 @@
...
+		std::cout << "Invalid Option, please enter a valid value " << std::endl;
+		std::cin >> UserInput;
+	};
+	return UserInput;
+}
+
+//-----------------------------------------------------------------------------
+void FMenu::ComeBackMainManu()
+{
+	while (std::cin.bad())
```
JDAlzate, line 367: Why use this condition for your loop?

Array_Vector_Exercises/Exercise1/TStaticArray.h
```diff
@@ -0,0 +1,94 @@
...
+	};
+
+	//----------------------------------------------------------
+	int GetSize()
+	{
+		return SizeArray;
+	};
+
+	//----------------------------------------------------------
+	void Fill(T Type)
```
JDAlzate, line 68: Why are you passing this in as a copy?

Array_Vector_Exercises/Exercise1/TStaticArray.h
JDAlzate: Yep

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
+#pragma once
+
+template<typename T>
+class TDynamicArray
+{
+private:
+	int SizeArray;
```
JDAlzate, line 7: Remember that it's a bit redundant to call all your variables Array. Size and Capacity would be more than enough since their context is very clear. This reduces noise in the implementations.

Array_Vector_Exercises/Exercise1/TDinamicArray.h
JDAlzate: Yep, that would not be good, you wouldn't be left with any array.

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+		SizeArray = 0;
+	}
+
+	//--------------------------------------------------------------------------
+	void PushBack(const T& InType)
+	{
+		if (SizeArray == CapacityArray)
+		{
+			Reserve(CapacityArray * 2);
+			DynamicArray[SizeArray] = InType;
```
JDAlzate, line 145: There's duplicate code here. Why write this twice?
You can do the reserve only if the condition is met, but have the rest of the code that always needs to execute outside of the if.
```
if (SizeArray == CapacityArray)
{
    Reserve(CapacityArray * 2);
}

DynamicArray[SizeArray] = InType;
SizeArray++
```

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+		if (Index < SizeArray && Index >= 0)
+		{
+			if (SizeArray == CapacityArray)
+			{
+				Reserve(CapacityArray * 2);
+			}
+				
+			for (int i = SizeArray; i > Index; i--)
+			{
+				DynamicArray[Index] = DynamicArray[i - 1];
```
JDAlzate, line 176: This doesn't seem like it would work. You changing the same value every iteration.

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+			SizeArray++;	
+		}
+	}
+
+	//-----------------------------------------------------------------------------
+	void Erase(const int Index)
+	{
+		for (int i = 0; i < SizeArray - 1; i++)
+		{
+			DynamicArray[Index] = DynamicArray[Index + 1];
```
JDAlzate, line 188: Same here, doesn't seem like this function would work at all.

Array_Vector_Exercises/Exercise1/TDinamicArray.h
JDAlzate: Yep. What if NewSize is greater than the current capacity.

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+	{
+		Reserve(AppendArray.GetSize() + SizeArray);
+		for (int i = 0; i < AppendArray.GetSize(); i++)
+		{
+			PushBack(AppendArray[i]);
+		}
+	}
+
+	//----------------------------------------------------------------------------
+	void Replace(const int Index, T& InType)
```
JDAlzate, line 210: This function is a bit too simple, I think. It is unnecessary because you are literally doing DynamicArray[Index] = InType, which the user could just do themselves.

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+	//---------------------------------------------------------------------------
+	void SwapIndex(const int Index1, const int Index2)
+	{
+		T SwapType = DynamicArray[Index1];
+		DynamicArray[Index1] = DynamicArray[Index2];
+		DynamicArray[Index2] = SwapType;		
+	}
+
+	//----------------------------------------------------------------------------
+	void Split(TDynamicArray<T>& Array1, TDynamicArray<T>& Array2, const int Index)
```
JDAlzate, line 224: Remember that OutParameters use the prefix `Out`

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+	{
+		if (!Array1.IsEmpty() && !Array2.IsEmpty())
```
JDAlzate, line 226: OutParameters are expected to come in empty. This is nice, but not really necessary, if there's an error because the arrays are not empty, then that's on the user, not on you.

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+		{
+			Array1.Clear();
+			Array2.Clear();
+		}
+		for (int i = 0; i < Index; i++)
+		{
+			Array1.Reserve(Index);
```
JDAlzate, line 233: Why are you reserving the same value on every iteration?

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+			Array1.PushBack(DynamicArray[i]);
+		}
+		for (int i = Index; i < CapacityArray; i++)
```
JDAlzate, line 236: Why are you copying the capacity elements that are not within the size as well?

Array_Vector_Exercises/Exercise1/TDinamicArray.h
```diff
@@ -0,0 +1,242 @@
...
+		{
+			Array1.Reserve(CapacityArray - Index);
```
JDAlzate, line 238: You're not doing anything in this line. You're reserving to the first array. Also, why would this be done in every iteration.

JDAlzate (changes requested):
Review is up!

PR #14, opened 2022-09-19
Linked list.

simonkytr:
these are the exercises for the Link List part, using Lambdas, and for each
image

### DIFF
--- a/LinkedList/LinkedList/DoublyList.h
+++ b/LinkedList/LinkedList/DoublyList.h
@@ -10,7 +10,7 @@ private:
 	{
 		Type Data;
 		FNode* Next = nullptr;
-		FNode* Previuos = nullptr;
+		FNode* previous = nullptr;
 	};
 	FNode* Head = nullptr;
 	FNode* Tail = nullptr;
@@ -103,7 +103,7 @@ public:
 			Current = Tail;
 			for (int i = Size - 1; i < Index / 2; i--)
 			{
-				Current = Current->Previuos;
+				Current = Current->previous;
 			}
 			return Current->Data;
 		}
@@ -115,7 +115,7 @@ public:
 		if (Index < Size / 2)
 		{
 			Current = Head;
-			for (int i = 0; i < Index / 2; i++)
+			for (int i = 0; i < Index; i++)
 			{
 				Current = Current->Next;
 			}
@@ -124,9 +124,9 @@ public:
 		else
 		{
 			Current = Tail;
-			for (int i = Size - 1; i < Index / 2; i--)
+			for (int i = Size - 1; i > Index; i--)
 			{
-				Current = Current->Previuos;
+				Current = Current->previous;
 			}
 			return Current->Data;
 		}
@@ -186,15 +186,15 @@ public:
 		FNode* NewHead = new FNode;
 		NewHead->Data = InHead;
 
-		if (Head == nullptr)
+		if (!Head)
 		{
 			Tail = NewHead;
 		}
 		else
 		{
 			NewHead->Next = Head;
-			Head->Previuos = NewHead;
-			NewHead->Previuos = nullptr;
+			Head->previous = NewHead;
+			NewHead->previous = nullptr;
 		}
 		Head = NewHead;
 		Size++;
@@ -203,7 +203,7 @@ public:
 	//--------------------------------------------------
 	void AddTail(const Type& InTail)
 	{
-		if (Head == nullptr)
+		if (IsEmpty())
 		{
 			AddHead(InTail);
 		}
@@ -211,7 +211,7 @@ public:
 		{
 			Tail->Next = new FNode;
 			Tail->Next->Data = InTail;
-			Tail->Next->Previuos = Tail;
+			Tail->Next->previous = Tail;
 			Tail = Tail->Next;
 			
 			Size++;
@@ -231,8 +231,7 @@ public:
 		}
 		else
 		{
-			FNode* Current = GetNode(Index);
-			if (Current == Tail)
+			if (Index == Size)
 			{
 				AddTail(NewData);
 			}
@@ -241,10 +240,12 @@ public:
 				FNode* NewNode = new FNode;
 				NewNode->Data = NewData;
 
-				NewNode->Previuos = Current->Previuos;
-				Current->Previuos->Next = NewNode;
+				FNode* Current = GetNode(Index);
+				NewNode->previous = Current->previous;
+				Current->previous->Next = NewNode;
+
 				NewNode->Next = Current;
-				Current->Previuos = NewNode;	
+				Current->previous = NewNode;	
 
 				Size++;
 			}
@@ -284,11 +285,11 @@ public:
 		{
 			FNode* Current = GetNode(Index);
 			
-			Current->Previuos->Next = Current->Next;
+			Current->previous->Next = Current->Next;
 
 			if (Current->Next != nullptr)
 			{
-				Current->Next->Previuos = Current->Previuos;
+				Current->Next->previous = Current->previous;
 			}
 			
 			delete Current;
@@ -397,7 +398,7 @@ public:
 				Current = Tail;
 				for (int i = Size - 1; i > Index; i--)
 				{
-					Current = Current->Previuos;
+					Current = Current->previous;
 				}
 				return Current;
 			}
@@ -409,20 +410,20 @@ public:
 			if (Index < Size / 2)
 			{
 				Current = Head;
-				for (int i = 0; i < Index / 2; i++)
+				for (int i = 0; i < Index; i++)
 				{
 					Current = Current->Next;
 				}
-				return Current;
+				return Current->Data;
 			}
 			else
 			{
 				Current = Tail;
-				for (int i = Size - 1; i < Index / 2; i--)
+				for (int i = Size - 1; i > Index; i--)
 				{
-					Current = Current->Previuos;
+					Current = Current->previous;
 				}
-				return Current;
+				return Current->Data;
 			}
 		}
 
--- a/LinkedList/LinkedList/DoublyList.h
+++ b/LinkedList/LinkedList/DoublyList.h
@@ -1,0 +1,429 @@
+#pragma once
+#include <functional>
+
+template <typename Type>
+class TDLList
+{
+private:
+	int Size = 0;
+	struct FNode
+	{
+		Type Data;
+		FNode* Next = nullptr;
+		FNode* Previuos = nullptr;
+	};
+	FNode* Head = nullptr;
+	FNode* Tail = nullptr;
+
+public:
+	TDLList() {}
+
+	~TDLList() 
+	{
+		Clear();
+	}
+
+	//-----------------------------------CopyConstructor
+	TDLList(const TDLList<Type>& OtherList)
+	{
+		Clear();
+		FNode* CurrentNode = OtherList.Head;
+
+		while(CurrentNode)
+		{
+			AddTail(CurrentNode->Data);
+			CurrentNode = CurrentNode->Next;
+		}
+	}
+
+	//------------------------------------Iterator
+	class FIterator
+	{
+	private:
+		FNode* Current;
+	public:
+		FIterator(FNode* NewCurrent)
+		{
+			Current = NewCurrent;
+		}
+
+		FIterator& operator++()
+		{
+			Current = Current->Next;
+			return *this;
+		}
+
+		Type& operator*()
+		{
+			return Current->Data;
+		}
+
+		const Type& operator*() const
+		{
+			return Current->Data;
+		}
+
+		bool operator!=(const FIterator& Other)
+		{
+			return Current != Other.Current;
+		}
+
+		bool operator!=(const FIterator& Other) const
+		{
+			return Current != Other.Current;
+		}
+	};
+
+	FIterator begin()
+	{
+		return FIterator(Head);
+	}
+
+	FIterator end()
+	{
+		return FIterator(nullptr);
+	}
+	
+
+	//---------------------------------Access
+	Type& operator[](const int Index)
+	{
+		FNode* Current;
+		if (Index < Size / 2)
+		{
+			Current = Head;
+			for (int i = 0; i < Index / 2; i++)
+			{
+				Current = Current->Next;
+			}
+			return Current->Data;
+		}
+		else
+		{
+			Current = Tail;
+			for (int i = Size - 1; i < Index / 2; i--)
+			{
+				Current = Current->Previuos;
+			}
+			return Current->Data;
+		}
+	}
+
+	const Type& operator[](const int Index) const
+	{
+		FNode* Current;
+		if (Index < Size / 2)
+		{
+			Current = Head;
+			for (int i = 0; i < Index / 2; i++)
+			{
+				Current = Current->Next;
+			}
+			return Current->Data;
+		}
+		else
+		{
+			Current = Tail;
+			for (int i = Size - 1; i < Index / 2; i--)
+			{
+				Current = Current->Previuos;
+			}
+			return Current->Data;
+		}
+	}
+
+	//-------------------------------------------------------
+	TDLList& operator=(const TDLList<Type>& OtherList)
+	{
+		Clear();
+		FNode* CurrentNode = OtherList.Head;
+
+		for (int i = 0; i < OtherList.Size; i++)
+		{
+			AddTail(CurrentNode->Data);
+			CurrentNode = CurrentNode->Next;
+		}
+		return *this;
+	}
+
+	//--------------------------------------------------------
+	Type& GetHead()
+	{
+		return Head->Data;
+	}
+
+	const Type& GetHead() const
+	{
+		return Head->Data;
+	}
+
+	//---------------------------------------------------------
+	Type& GetTail()
+	{
+		return Tail->Data;
+	}
+
+	const Type& GetTail() const
+	{
+		return Tail->Data;
+	}
+
+	//---------------------------------Size
+	bool IsEmpty() const
+	{
+		return Size == 0;
+	}
+
+	//------------------------------------------------
+	int GetSize() const
+	{
+		return Size;
+	}
+
+	//-----------------------------------Modifiers
+	void AddHead(const Type& InHead)
+	{
+		FNode* NewHead = new FNode;
+		NewHead->Data = InHead;
+
+		if (Head == nullptr)
+		{
+			Tail = NewHead;
+		}
+		else
+		{
+			NewHead->Next = Head;
+			Head->Previuos = NewHead;
+			NewHead->Previuos = nullptr;
+		}
+		Head = NewHead;
+		Size++;
+	}
+
+	//--------------------------------------------------
+	void AddTail(const Type& InTail)
+	{
+		if (Head == nullptr)
+		{
+			AddHead(InTail);
+		}
+		else
+		{
+			Tail->Next = new FNode;
+			Tail->Next->Data = InTail;
+			Tail->Next->Previuos = Tail;
+			Tail = Tail->Next;
+			
+			Size++;
+		}
+	}
+
+	//--------------------------------------------------
+	void Insert(const Type& NewData, const int Index)
+	{
+		if (Index < 0 || Index > Size)
+		{
+			return;
+		}
+		else if (Index == 0)
+		{
+			AddHead(NewData);
+		}
+		else
+		{
+			FNode* Current = GetNode(Index);
+			if (Current == Tail)
+			{
+				AddTail(NewData);
+			}
+			else
+			{
+				FNode* NewNode = new FNode;
+				NewNode->Data = NewData;
+
+				NewNode->Previuos = Current->Previuos;
+				Current->Previuos->Next = NewNode;
+				NewNode->Next = Current;
+				Current->Previuos = NewNode;	
+
+				Size++;
+			}
+		}
+	}
+
+	//-------------------------------------------------------
+	void RemoveHead()
+	{
+		if (Size != 0)
+		{
+			FNode* NewNode = Head->Next;
+			delete Head;
+			Head = NewNode;
+
+			if (Size == 1)
+			{
+				Head = nullptr;
+				Tail = nullptr;
+			}
+			--Size;
+		}
+	}
+
+	//--------------------------------------------------------
+	void Remove(const int Index)
+	{
+		if (Index < 0 || Index > Size)
+		{
+			return;
+		}
+		if (Index == 0)
+		{
+			RemoveHead();
+		}
+		else
+		{
+			FNode* Current = GetNode(Index);
+			
+			Current->Previuos->Next = Current->Next;
+
+			if (Current->Next != nullptr)
+			{
+				Current->Next->Previuos = Current->Previuos;
+			}
+			
+			delete Current;
+			Size--;
+		}
+	}
+
+	//-----------------------------------------------------------
+	void Clear()
+	{
+		while (Head)
+		{
+			RemoveHead();
+		}
+	}
+
+	//----------------------------------------ForEach
+	template<typename Pred>
+	void ForEach(const Pred& Predicate)
+	{
+		for (Type& Temporal : *this)
+		{
+			Predicate(Temporal);
+		}
+	}
+
+	//----------------------------------------Lambdas
+	template< typename Pred>
+	Type* FindByPredicate(const Pred& Predicate) const
+	{
+		for (Type& Current : *this)
+		{
+			if (Predicate(Current))
+			{
+				return &Current->Data;
+			}
+		}
+		return nullptr;
+	};
+
+	//-------------------------------------------------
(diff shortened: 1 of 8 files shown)


LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,352 @@
...
+		return FIterator(nullptr);
+	}
+
+	FIterator end()
+	{
+		return FIterator(nullptr);
+	}
+
+	//-------------------------------CopyConstructor
+	TSLList(const TSLList<Type>& OtherTSLlist)
```
GaxysGalactic: I think in terms of organization, the copy constructor (and possibly the operator= override) should be above the Iterator class, unless you are going to be using the Iterator in the method (which you could)

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	~TSLList()
+	{
+		Clear();
+	}
+
+	//--------------------------------FIterator
+	class FIterator
+	{
```
GaxysGalactic, line 41: For encapsulation, you can maybe do:
- Public section for members (Iterator could go here)
- Private section for members 
- Public section for methods
- Private section for methods (If you have any)
Though this is merely a suggestion. I'm not entirely sure as to the proper placement of Iterators and iterator methods in a given class.

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,352 @@
...
+	//-------------------------------CopyConstructor
+	TSLList(const TSLList<Type>& OtherTSLlist)
+	{
+		FNode* Temporal = OtherTSLlist.Head;
+
+		for (int i = 0; i < Size; i++)
+		{
+			AddTail(Temporal->Element);
+			Temporal = Temporal->Next;
+		}
```
GaxysGalactic: David noted that perhaps now with the power of iterators, we could change functions like these ones to use them. Of course, it isn't necessary, but if you have the time, try to implement it in one of these to see how it works.

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,352 @@
...
+		bool operator!=(const FIterator& Other) const
+		{
+			return Current != Other.Current;
+		}
+	};
+
+	FIterator begin()
+	{
+		if(Head)
+			return FIterator(Head);
```
GaxysGalactic, line 81: This line might not be that necessary, as by default, according to your member initialization at the top of the class, Head should start as nullptr. Therefore, lines 70 and 71 would be doing the same either way.

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,352 @@
...
+
+	FIterator end()
+	{
+		return FIterator(nullptr);
+	}
+
+	//-------------------------------CopyConstructor
+	TSLList(const TSLList<Type>& OtherTSLlist)
+	{
+		FNode* Temporal = OtherTSLlist.Head;
```
GaxysGalactic: Hmm, I think this variable might be understood a lot better if it were named something like _Current_, for example. Temporal (which means related to time in English) / Temporary might be a bit confusing, as what it is pointing to is a more "permanent" node in the list as opposed to a temporary created one.

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,352 @@
...
+			Head = NewHead;
+			Tail = NewHead;
+			Size++;
+		}
+		else
+		{
+			NewHead->Next = Head;
+			Head = NewHead;
+			Size++;
+		}
```
GaxysGalactic: This could perhaps be made more efficient in terms of line use. Notice that lines 169 and line 176 are functionally doing the same thing, and `NewHead->Next = Head` could still be used in the initial scenario, being harmless. What does need to change, though, is that before you apply that, you use the if condition you write:
```
if (Head == nullptr)
{
     Tail = NewHead;
}
```

As that's the main difference between the two cases.

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,352 @@
...
+
+	//-------------------------------------------------
+	void AddTail(const Type& InTail)
+	{
+
+		if (Head == nullptr)
+		{
+			AddHead(InTail);
+		}
+
```
GaxysGalactic: Very nitpicky but this space between the if and the else statement might be a bit confusing at first sight. This would be slightly more readable by removing line 189.

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,352 @@
...
+			Tail = nullptr;
+			--Size;
+		}
+	}
+
+	//---------------------------------------------------------
+	void Remove(const int Index)
+	{
+		FNode* Temporal = Head;
+
```
GaxysGalactic: Given that you created a RemoveHead() function, you could place an if statement to call it if the Index == 0 here. Though it depends on how you want your LinkedList to function: maybe according to your implementation, the proper way to remove the head is to call the RemoveHead() function only.

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,352 @@
...
+			Head = NextItem;
+		}
+		Size = 0;
+		Head = nullptr;
+		Tail = nullptr;
+	}
+
+	//----------------------------------------ForEach
+	template<typename Function>
+	void forEach(const Function& SomeFunction)
```
GaxysGalactic: I also initially used SomeFunction here, as that's what JD used during the lecture. However, on Discord he posted the updated version of the signature, and ideally it should be called "Pred" for the typename and "Predicate" for the parameter name, like you have below.

Also, the F in ForEach should be capitalized, I think.

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,352 @@
...
+		for (int i = 0; i < Size; i++)
+		{
+			SomeFunction(Temporal);
+			Temporal = Temporal->Next;
+		}
+	}
+
+	//----------------------------------------Lambdas
+	template< typename Pred>
+	Type* FindByPredicate(const Pred& Predictate)
```
GaxysGalactic: Small typo here, though not too serious as its just on the parameter name. Predicate instead of PredicTate.

GaxysGalactic (changes requested):
Great job! There are a few small suggestions which you could try to improve the readability and possible efficiency of the code, but thankfully I don't see any errors when it comes to proper functionality.

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+#include <functional>
+
+template <typename Type>
+class TSLList
+{
+private:
+	int Size;
+	struct FNode
+	{
+		Type Element;
```
DvdHalliday: I told Sergio this too, I don't actually feel like Element is a correct name for what's inside the node, I mean, a node might be an element, but what does it actually mean to ask the element of an element? Santiago googled it and the most common name for it would be Data, and that does make more sense as you're asking an element what its data is.

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		Size = 0;
+	}
+
+	~TSLList()
+	{
+		Clear();
+	}
+
+	//-------------------------------CopyConstructor
+	TSLList(const TSLList<Type>& OtherTSLlist)
```
DvdHalliday: No need to call it OtherTSLlist.. when you already know it is a TSLList

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		Type Element;
+		FNode* Next = nullptr;
+	};
+	FNode* Head = nullptr;
+	FNode* Tail = nullptr;
+
+public:
+	TSLList()
+	{
+		Size = 0;
```
DvdHalliday: Size should be initialised on declaration

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+	{
+		Clear();
+	}
+
+	//-------------------------------CopyConstructor
+	TSLList(const TSLList<Type>& OtherTSLlist)
+	{
+		FNode* Temporal = OtherTSLlist.Head;
+
+		for (int i = 0; i < Size; i++)
```
DvdHalliday: Ok, so discussing it with Santiago, there are actually two problems here. First, since you're not initialising Size on declaration, as it is a built-in type it will become whatever memory used to occupy its place. So, for all we know Size might be 8'000.000, so you're potentially trying to do something eight million times and this might also call trouble since you're likely soon to set your temporal variable to nullptr, so it should give you that error too.

DvdHalliday: The second problem has to do with the fact that you want to iterate untill i reaches OtherTSLlist.Size, not the current Size, which should be 0.

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		return FIterator(Head);
+	}
+
+	FIterator end()
+	{
+		return FIterator(nullptr);
+	}
+
+	//-------------------------Access
+	Type& operator[] (const int Index)
```
DvdHalliday: Extra space after the []

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		FNode* Temporal = Head;
+		for (int i = 0; i < Index; i++)
+		{
+			Temporal = Temporal->Next;
+		}
+		return Temporal->Element;
+	}
+
+	//-----------------------------------------------
+	TSLList& operator= (const TSLList<Type>& OtherTSLlist)
```
DvdHalliday: Same about the space after the = and the thing about OtherTSLlist's name

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+	{
+		Clear();
+		FNode* CurrentNode = OtherTSLlist.Head;
+
+		for (int i = 0; i < Size; i++)
```
DvdHalliday: Also, same thing about using this object's Size

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	//---------------------------------Size
+	bool IsEmpty()
+	{
+		return Size == 0;
+	}
+
+	//------------------------------------------------
+	int GetSize()
```
DvdHalliday: this should be a const function

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		return Tail->Element;
+	}
+
+	const Type& GetTail() const
+	{
+		return Tail->Element;
+	}
+
+	//---------------------------------Size
+	bool IsEmpty()
```
DvdHalliday: const function

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		{
+			Head = NewHead;
+			Tail = NewHead;
+			Size++;
+		}
+		else
+		{
+			NewHead->Next = Head;
+			Head = NewHead;
+			Size++;
```
DvdHalliday: since you're not returning on your function, and the last instruction is the same for both of them, you could simply do the size++ outside the if-else statement

DvdHalliday: Actually you could do the same with the Head = NewHead; instruction, or even before the statement

LinkedList/LinkedList/TSLList.h
DvdHalliday: yeah, i think this is it

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		}
+	}
+
+	//-------------------------------------------------
+	void AddTail(const Type& InTail)
+	{
+
```
DvdHalliday: Unnecessary blank space

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		if (Index == 0)
+		{
+			AddHead(NewItem);
+		}
+		else if(Index == Size)
+		{
+			AddTail(NewItem);
+		}
+
+		else if (Index > 0 && Index < Size)
```
DvdHalliday: I think to check on bounds first might be best, that way you could simply exit the function on such cases, else you will need to check 3 conditions before exiting

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		{
+			FNode* Temporal = Head;
+			FNode* Container = new FNode;
+
+			for (int i = 0; i < Index; i++)
+			{
+				Temporal = Temporal->Next;
+			}
+			Container->Element = Temporal->Element;
```
DvdHalliday: declare Container after the for loop, there is no need to declare it before if you're not gonna' use it, it's confusing

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+
+			Temporal->Element = NewItem;
+			Container->Next = Temporal->Next;
+			Temporal->Next = Container;
+			Size++;
```
DvdHalliday: There is a way of doing all of this with just one auxiliar node, using more auxiliar variables than you need might interfere with readability

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+	{
+		if (Size != 0)
+		{
+			FNode* NewHead = Head->Next;
+			delete Head;
+			Head = NewHead;
+			--Size;
+		}
+
+		else if (Size == 1)
```
DvdHalliday: This condition is never met, if Size is 1, then it is different to 0, so the if scope will run and that's the end of it, so you will have a Tail pointing to memory that should give you an error

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	//---------------------------------------------------------
+	void Remove(const int Index)
+	{
+		FNode* Temporal = Head;
+
+		if (Index == 0)
+		{
+			RemoveHead();
```
DvdHalliday, line 248: Nice!

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		}
+
+		else if (Index >= 0 && Index <= Size)
```
DvdHalliday: This could be Index > 0 since you're already considering that case

DvdHalliday: Also, same thing about indexes outta' bounds, it'd be nice if that was the first consideration of your function

DvdHalliday: It would also help with some readability. For even though else ifs work, this is not the best use for them. On conditions like these if-else statements are good enough

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+			Head = nullptr;
+			Tail = nullptr;
+			--Size;
+		}
+	}
+
+	//---------------------------------------------------------
+	void Remove(const int Index)
+	{
+		FNode* Temporal = Head;
```
DvdHalliday: Temporal should be declared in the scope you're going to use it, so about in line 257

LinkedList/LinkedList/TSLList.h
DvdHalliday: Another thing, Size should be out of bounds here

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+
+		else if (Index >= 0 && Index <= Size)
+		{
+			for (int i = 0; i < Index - 1; i++)
+			{
+				Temporal = Temporal->Next;
+			}
+
+			FNode* Current = Temporal->Next;
+			Temporal->Next = Current->Next;
```
DvdHalliday, line 254: If your index is Size, then Current is going to be set to nullptr, so the moment you run this line you're dereferencing nullptr, which should crash

DvdHalliday, line 254: Also, I think you could do all of this with a single variable, but then again that's simply my opinion

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		FNode* NextItem;
+
+		for (int i = 0; i < Size; i++)
+		{
+			NextItem = Head->Next;
+			delete Head;
+			Head = NextItem;
+		}
+		Size = 0;
+		Head = nullptr;
```
DvdHalliday: Head should already be set to nullptr if everything went down correctly

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		Tail = nullptr;
+	}
+
+	//----------------------------------------ForEach
+	template<typename Pred>
+	void forEach(const Pred& Predicate)
```
DvdHalliday: F should be capitalised

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+			{
+				return &Temporal->Element;
+			}
+		}
+		return nullptr;
+	};
+
+	//-------------------------------------------------
+	template< typename Pred>
+	TSLList<Type> FilterByPredicate(const Pred& Predicate)
```
DvdHalliday: const function

LinkedList/LinkedList/TSLList.h
DvdHalliday: There's still a typo, it's Predicate

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+	{
+		FNode* Temporal;
+		TSLList <Type> Filterlist;
```
DvdHalliday: extra space after TSLList

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+			}
+		}
+		return nullptr;
+	};
+
+	//-------------------------------------------------
+	template< typename Pred>
+	TSLList<Type> FilterByPredicate(const Pred& Predicate)
+	{
+		FNode* Temporal;
```
DvdHalliday: No need for this variable

LinkedList/LinkedList/TSLList.h
DvdHalliday: The l in List should be capitalised

LinkedList/LinkedList/TSLList.h
DvdHalliday: Also, const function

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+		for (int i = 0; i < Size; i++)
+		{
+			if (Predictate(Temporal))
+			{
+				Remove(i);
+				i--;
+			}
+			Temporal = Temporal->Next;
+		}
+		return false;
```
DvdHalliday: This function has a return type set to void, so returning a bool doesn't do anything and actually makes no sense

LinkedList/LinkedList/TSLList.h
```diff
@@ -0,0 +1,350 @@
...
+	};
+
+	//-------------------------------------------------
+	template< typename Pred>
+	void RemoveAllByPredicate(const Pred& Predicate)
+	{
+		FNode* Temporal = Head;
+		for (int i = 0; i < Size; i++)
+		{
+			if (Predictate(Temporal))
```
DvdHalliday: Careful, parameter's name is Predicate but you're calling a Predictate function here

DvdHalliday (changes requested):
Well done,
I hope I filled in some gaps that escaped Santiago's eyes!

LinkedList/LinkedList/Queue.h
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	void Dequeue()
+	{
+		QueueArray.RemoveHead();
+	}
+
+	void Front()
+	{
+
```
GaxysGalactic: Oh noes, I think the formatting here turned a little weird. Check on your project file since at least here on GitHub, some random spacing is here between the brackets.

LinkedList/LinkedList/Queue.h
```diff
@@ -0,0 +1,30 @@
...
+	{
+		QueueArray.AddTail(NewData);
+	}
+
+	void Dequeue()
+	{
+		QueueArray.RemoveHead();
+	}
+
+	void Front()
```
GaxysGalactic: Front shouldn't be a void function, it should return something, much like Stack's Top function. You should probably add a const and non-const version as well.

LinkedList/LinkedList/TSLList.h
```diff
@@ -5,10 +5,10 @@ template <typename Type>
 class TSLList
 {
 private:
-	int Size;
+	int Size = 0;
 	struct FNode
 	{
-		Type Element;
+		Type Data;
```
GaxysGalactic, line 11: Nice change with changing Element to Data. I did so for my doubly linked list as I think I prefer that term a little more now.

LinkedList/LinkedList/TSLList.h
```diff
@@ -339,14 +320,33 @@ class TSLList
...
+		FNode* GetNode(const int Index)
+		{
+			FNode* Temporal = Head;
+			for (int i = 0; i < Index - 1; i++)
+			{
+				Temporal = Temporal->Next;
+			}
+			return Temporal;
+		}
+
```
GaxysGalactic, line 342: Good job on the GetNode private function! It is an easy way to avoid rewriting code, and you could inmediately change a lot of functions to use the iterator now by simply changing this method.

LinkedList/LinkedList/DoublyList.h
```diff
@@ -0,0 +1,429 @@
...
+template <typename Type>
+class TDLList
+{
+private:
+	int Size = 0;
+	struct FNode
+	{
+		Type Data;
+		FNode* Next = nullptr;
+		FNode* Previuos = nullptr;
```
GaxysGalactic: Small typo, but not the worst in the world, it should be PreviOUs

LinkedList/LinkedList/DoublyList.h
```diff
@@ -0,0 +1,429 @@
...
+	
+
+	//---------------------------------Access
+	Type& operator[](const int Index)
+	{
+		FNode* Current;
+		if (Index < Size / 2)
+		{
+			Current = Head;
+			for (int i = 0; i < Index / 2; i++)
```
GaxysGalactic, line 95: Careful! Why should it go to half of its Index? The half part of the algorithm was already dealt with (starting point if conditional on line 92). This might not bring what you want if you tried to access [4] in an array of 12 elements, for example. It would enter into this for loop, but would only iterate twice (as Index / 2 is 2)

LinkedList/LinkedList/DoublyList.h
```diff
@@ -0,0 +1,429 @@
...
+		return Size;
+	}
+
+	//-----------------------------------Modifiers
+	void AddHead(const Type& InHead)
+	{
+		FNode* NewHead = new FNode;
+		NewHead->Data = InHead;
+
+		if (Head == nullptr)
```
GaxysGalactic: Remember that you could've switched this around and either changed the inside to say `!Head` or switch the if and else contents and use `Head` here

LinkedList/LinkedList/DoublyList.h
```diff
@@ -0,0 +1,429 @@
...
+			NewHead->Previuos = nullptr;
+		}
+		Head = NewHead;
+		Size++;
+	}
+
+	//--------------------------------------------------
+	void AddTail(const Type& InTail)
+	{
+		if (Head == nullptr)
```
GaxysGalactic: Same possible change here as the one suggested in the previous function.

LinkedList/LinkedList/DoublyList.h
```diff
@@ -0,0 +1,429 @@
...
+		{
+			AddHead(NewData);
+		}
+		else
+		{
+			FNode* Current = GetNode(Index);
+			if (Current == Tail)
+			{
+				AddTail(NewData);
+			}
```
GaxysGalactic: This could've been placed on another if statement after the Index one, maybe one being `Index == Size`, this might help improve readability and avoid the conditional here.

GaxysGalactic (changes requested):
Very good! A few comments, as well as maybe a suggestion or two on what you can change.

